PR #6, opened 2020-04-24
Format code with goimports

mrKisaoLamb:
- reformat code with goimports
- add info about goimports into README


mpryc:
I think it's ok to submit, however better to add validation test first to ensure it passes. This way code formatting will be preventing from other merges in the future.

e.g. job that does something as simple as:
goimports -d .

mrKisaoLamb:
> I think it's ok to submit, however better to add validation test first to ensure it passes. 
Agree
> e.g. job that does something as simple as:
> goimports -d .
yeah this code is result of goimports (this patch and other one for nova was the reasone why I go to you to add checks ;) )

PR #11, opened 2020-05-21
Add ovn support

mrKisaoLamb:


deploy/role.yaml
```diff
@@ -41,15 +48,44 @@ rules:
 - apiGroups:
   - apps
   resourceNames:
-  - neutron-operator
+  - nova-operator
```
bogdando, line 51: typo?

pkg/controller/ovncontroller/ovncontroller_controller.go
```diff
@@ -0,0 +1,311 @@
...
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileOVNController) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling OVNController")
+
+	// Fetch the OVNController instance
+	instance := &neutronv1.OVNController{}
+	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
+	if err != nil {
+		if errors.IsNotFound(err) {
```
bogdando, line 110: nit: no need to nest that :)

pkg/ovncontroller/configmap.go
```diff
@@ -0,0 +1,51 @@
...
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "ConfigMap",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cmName,
+			Namespace: cr.Namespace,
+		},
+		Data: map[string]string{
+			"temp": "temp",
```
dprince, line 46: so this configmap is a placeholder and this temp:temp mapping will go away once you move the neutron conf over?

pkg/ovncontroller/configmap.go
```diff
@@ -0,0 +1,51 @@
...
+		},
+		Data: map[string]string{
+			"ovn.sh": util.ExecuteTemplateFile(strings.ToLower(cr.Kind)+"/ovn.sh", nil),
+		},
+	}
+
+	return cm
+}
+
+// TemplatesConfigMap - custom nova config map
```
dprince, line 32: neutron right?

pkg/controller/ovsnodeosp/ovsnodeosp_controller.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	// Watch for changes to secondary resource Pods and requeue the owner OVSNodeOsp
+	err = c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &neutronv1.OVSNodeOsp{},
+	})
+	if err != nil {
+		return err
+	}
```
dprince, line 74: Perhaps missing here do we need to watch for changes to the DaemonSet's too right?

pkg/controller/ovsnodeosp/ovsnodeosp_controller.go
```diff
@@ -0,0 +1,323 @@
...
+				},
+			},
+			InitialDelaySeconds: 15,
+			PeriodSeconds:       5,
+		},
+		Env: []corev1.EnvVar{
+			{
+				Name:  "TEMPLATES_CONFIG_HASH",
+				Value: templatesConfigHash,
+			},
```
stuggi, line 289: should we also add the SCRIPTS_CONFIG_HASH here to trigger restart when the scripts get updated, or maybe we don't want this behavior for the ovs-node-osp?

pkg/ovncontroller/volumes.go
```diff
@@ -0,0 +1,117 @@
...
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: "/var/lib/openvswitch",
+				},
+			},
+		},
+		{
+			Name: "env-overrides",
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: "/env",
```
stuggi, line 47: what can we control with the info in /env/${K8S_NODE}? I see it gets sourced in the script. Is it for debugging?

pkg/controller/ovsnodeosp/ovsnodeosp_controller.go
```diff
@@ -0,0 +1,323 @@
...
+			{
+				Name:  "OVS_LOG_LEVEL",
+				Value: "info",
```
stuggi, line 292: should a user be able to change this via the CR or a configmap entry?

pkg/controller/ovncontroller/ovncontroller_controller.go
```diff
@@ -0,0 +1,311 @@
...
+			Privileged: &trueVar,
+		},
+		Env: []corev1.EnvVar{
+			{
+				Name:  "TEMPLATES_CONFIG_HASH",
+				Value: templatesConfigHash,
+			},
+			{
+				Name:  "OVS_LOG_LEVEL",
+				Value: "info",
```
stuggi, line 280: should a user be able to change this via CR or configmap?

pkg/ovncontroller/configmap.go
mrKisaoLamb: Yes this is placeholder

README.md
```diff
@@ -23,6 +23,8 @@ Create CRDs
 
     oc create -f deploy/crds/neutron_v1_neutronovsagent_crd.yaml
     oc create -f deploy/crds/neutron_v1_neutronsriovagent_crd.yaml
+    oc create -f deploy/crds/neutron.openstack.org_ovsnodeosps_crd.yaml
```
luis5tb, line 26: you will need to create one or the other, right? perhaps worth to split it by ovs and ovn

deploy/crds/neutron.openstack.org_v1_ovncontroller_cr.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: neutron.openstack.org/v1
+kind: OVNController
+metadata:
+  name: ovncontroller
+  namespace: openstack
+spec:
+  ovnControllerImage: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:7d6ba1bcc4f403733e44cd2a008dafe1887501f7a3c74084566a4f4d1e46250d
```
luis5tb, line 7: shouldn't this be tripleO image?

deploy/crds/neutron.openstack.org_v1_ovsnodeosp_cr.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: neutron.openstack.org/v1
+kind: OVSNodeOsp
+metadata:
+  name: ovs-node-osp
+  namespace: openstack
+spec:
+  ovsNodeOspImage: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:7d6ba1bcc4f403733e44cd2a008dafe1887501f7a3c74084566a4f4d1e46250d
```
luis5tb, line 7: ditto

deploy/role.yaml
```diff
@@ -50,6 +57,32 @@ rules:
   - neutron.openstack.org
   resources:
   - '*'
```
luis5tb, line 59: should this be removed? we are listing the exact resources

pkg/controller/ovncontroller/ovncontroller_controller.go
```diff
@@ -0,0 +1,311 @@
...
+
+	// Watch for changes to primary resource OVNController
+	err = c.Watch(&source.Kind{Type: &neutronv1.OVNController{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
+	// Watch ConfigMaps owned by OVNController
+	err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForOwner{
+		IsController: false,
```
luis5tb, line 64: own but not controlled?

pkg/controller/ovncontroller/ovncontroller_controller.go
```diff
@@ -0,0 +1,311 @@
...
+				return reconcile.Result{}, err
+			}
+			r.setDaemonsetHash(instance, dsHash)
+			return reconcile.Result{RequeueAfter: time.Second * 10}, err
+		}
+	}
+
+	// Daemonset already exists - don't requeue
+	reqLogger.Info("Skip reconcile: Daemonset already exists", "Ds.Namespace", found.Namespace, "Ds.Name", found.Name)
+	return reconcile.Result{}, nil
```
luis5tb, line 212: perhaps we want to add a timing reconciliation here (similarly to the computenode operator), to ensure every X mins (5-10 mins) that everything needed is still there

pkg/controller/ovsnodeosp/ovsnodeosp_controller.go
```diff
@@ -0,0 +1,323 @@
...
+		err = r.client.Create(context.TODO(), scriptsConfigMap)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	} else if !reflect.DeepEqual(scriptsConfigMap.Data, foundScriptsConfigMap.Data) {
+		reqLogger.Info("Updating ScriptsConfigMap")
+		scriptsConfigMap.Data = foundScriptsConfigMap.Data
+	}
+
+	scriptsConfigMapHash, err := util.ObjectHash(scriptsConfigMap.Data)
```
luis5tb, line 129: is this used?

templates/ovsnodeosp/ovsnode.sh
```diff
@@ -0,0 +1,33 @@
...
+trap quit SIGTERM
+/usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --system-id=random
+ovs-appctl vlog/set "file:${OVS_LOG_LEVEL}"
+/usr/share/openvswitch/scripts/ovs-ctl --protocol=udp --dport=6081 enable-protocol
+
+sleep 5
+ovs-vsctl --may-exist add-br br-ex -- set Bridge br-ex fail-mode=secure
+ovs-vsctl --may-exist add-br br-isolated -- set Bridge br-isolated fail-mode=secure
+ovs-vsctl set open . external-ids:ovn-bridge=br-int
+ovs-vsctl set open . external-ids:ovn-remote=tcp:192.168.25.100:6642
```
luis5tb, line 25: that IP should not be hardcoded

luis5tb:
I left a few comments here and there. Also, this is missing the metadata which is needed

pkg/controller/ovncontroller/ovncontroller_controller.go
dprince: The compute-node-operator pattern is different though. If we add an explicit c.Watch on Daemonsets that would immediately trigger the operator to run if there are any changes (see review comment above). Periodically re-running the operator is good too but lets not rely on that for Daemonset change detection as those are a key component here.

pkg/controller/ovsnodeosp/ovsnodeosp_controller.go
dprince: See @stuggi's comment above. If we pass this into the env for the daemonset (or wherever it is applied) it will trigger an update there if a change is detected. Like if we update the scripts or something

pkg/controller/ovncontroller/ovncontroller_controller.go
stuggi: I think this is coming from some tests I did long time back on the nova-operator where I also set this. but I can not remember why I set it to false and why I have not switched it back. I agree, IsControlled: true should be ok, but as said I can not remember why I switched it to false there :(

README.md
mrKisaoLamb: We will remove ovs so will update README then

deploy/role.yaml
```diff
@@ -50,7 +57,32 @@ rules:
 - apiGroups:
   - neutron.openstack.org
   resources:
-  - '*'
+  - ovncontrollers
```
luis5tb, line 61: you need the "- '*'" here

deploy/scc.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: v1
+kind: SecurityContextConstraints
```
luis5tb, line 2: probably we just need a few of those, not all. But we can fine tune that in a follow up PR

pkg/apis/neutron/v1/ovsnnodeosp_types.go
```diff
@@ -0,0 +1,55 @@
...
+	// container image to run for the daemon
+	OvsNodeOspImage string `json:"ovsNodeOspImage"`
+	// service account used to create pods
+	ServiceAccount string `json:"serviceAccount"`
+	// Name of the worker role created for OSP computes
+	RoleName string `json:"roleName"`
+	// log level
+	OvsLogLevel string `json:"ovsLogLevel"`
+	// ovn south db ip
+	OvnSbRemote string `json:"ovnSbRemote"`
```
luis5tb, line 19: as discussed, this probably should be read from a configmap, instead of being part of the CRD, similarly to the approach @stuggi took for the nova-operator

templates/ovsnodeosp/ovsnode.sh
```diff
@@ -0,0 +1,33 @@
...
+ovs-appctl vlog/set "file:${OVS_LOG_LEVEL}"
+/usr/share/openvswitch/scripts/ovs-ctl --protocol=udp --dport=6081 enable-protocol
+
+sleep 5
+ovs-vsctl --may-exist add-br br-ex -- set Bridge br-ex fail-mode=secure
+ovs-vsctl --may-exist add-br br-isolated -- set Bridge br-isolated fail-mode=secure
+ovs-vsctl set open . external-ids:ovn-bridge=br-int
+ovs-vsctl set open . external-ids:ovn-remote=tcp:${OVN_SB_REMOTE}:6642
+ovs-vsctl set open . external-ids:ovn-encap-type=geneve
+export OVN_NODE_IP=`ip -4 -o addr show enp6s0 | awk '{print $4}' | cut -d"/" -f1`
```
luis5tb, line 27: this should not be hardcoded, probably part of the CRD for ovs

deploy/scc.yaml
mrKisaoLamb: agree will do in follow up patch

pkg/apis/neutron/v1/ovsnnodeosp_types.go
mrKisaoLamb: will be done in follow-up patch

deploy/crds/neutron.openstack.org_v1_ovncontroller_cr.yaml
dprince: > shouldn't this be tripleO image?

eventually we'd like OCP and OSP to use the same versions. But for now if we want them to match choosing the OCP version here is probably our only option. If the versions don't need to match (because we aren't testing any advanced bridging or anything) then using the RDO image would be better though

pkg/controller/ovncontroller/ovncontroller_controller.go
```diff
@@ -0,0 +1,315 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+var log = logf.Log.WithName("controller_ovncontroller")
+
+/**
+* USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
+* business logic.  Delete these comments after modifying this file.*
+ */
```
stuggi, line 35: we could remove this now

pkg/controller/ovncontroller/ovncontroller_controller.go
```diff
@@ -0,0 +1,315 @@
...
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the apiserver
+	client client.Client
+	scheme *runtime.Scheme
+}
+
+// Reconcile reads that state of the cluster for a OVNController object and makes changes based on the state read
+// and what is in the OVNController.Spec
+// TODO(user): Modify this Reconcile function to implement your Controller logic.  This example creates
+// a Pod as an example
```
stuggi, line 98: same, we should remove this

PR #14, opened 2020-07-03
Add OVNcontroller and OVSnodeOsp to cvs-generator

stuggi:
this adds new added OVNcontroller and OVSnodeOsp CRDs to the
csv-generator, also adds the scc generation as it is not handled
by olm correct.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -17,6 +17,7 @@ import (
 	"github.com/openstack-k8s-operators/neutron-operator/pkg/controller"
 	"github.com/openstack-k8s-operators/neutron-operator/version"
 
+	secv1 "github.com/openshift/api/security/v1"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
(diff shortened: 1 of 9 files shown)


cmd/manager/main.go
```diff
@@ -217,3 +236,66 @@ func serveCRMetrics(cfg *rest.Config, operatorNs string) error {
...
+		FSGroup: secv1.FSGroupStrategyOptions{
+			Type: secv1.FSGroupStrategyRunAsAny,
+		},
+		AllowPrivilegedContainer: true,
+		AllowPrivilegeEscalation: &trueVar,
+		AllowHostDirVolumePlugin: true,
+		AllowHostIPC:             true,
+		AllowHostNetwork:         true,
+		AllowHostPID:             true,
+		AllowHostPorts:           true,
```
dprince, line 266: we have this set to 'false' in deploy/scc.yaml.

stuggi: done set to false to have in sync

PR #17, opened 2020-07-15
Add scipts and improve Makefile to improve installation

mrKisaoLamb:


build_image.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+oc create -f deploy/crds/neutron_v1_neutronovsagent_crd.yaml
+oc create -f deploy/crds/neutron_v1_neutronsriovagent_crd.yaml
+oc create -f deploy/crds/neutron.openstack.org_ovsnodeosps_crd.yaml
+oc create -f deploy/crds/neutron.openstack.org_ovncontrollers_crd.yaml
+make
+operator-sdk build --image-builder buildah "${REGISTRY:-quay.io/openstack-k8s-operators/neutron-operator:v0.0.3}"
```
stuggi, line 7: maybe we should start have a latest tag?

dprince, line 7: I might make this variable more specific. REGISTRY isn't quite correct. Its more of a BUILD_IMAGE right?

Also, I do think we might want to be careful people don't run this and accidentally push to openstack-k8s-operators right? It might be better to just require it to be set.

mrKisaoLamb: Agree changed

mrKisaoLamb: Name changed. 
About this default, only teammates can push to quay.io/openstack-k8 so don't see a risk here, also you still need to set REGISTRY_AUTH

build_image.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+oc create -f deploy/crds/neutron_v1_neutronovsagent_crd.yaml
+oc create -f deploy/crds/neutron_v1_neutronsriovagent_crd.yaml
+oc create -f deploy/crds/neutron.openstack.org_ovsnodeosps_crd.yaml
+oc create -f deploy/crds/neutron.openstack.org_ovncontrollers_crd.yaml
+make
+operator-sdk build --image-builder buildah "${BUILD_IMAGE:-quay.io/openstack-k8s-operators/neutron-operator:latest}"
+sed -i 's|REPLACE_IMAGE|"${BUILD_IMAGE:-quay.io/openstack-k8s-operators/neutron-operator:latest}"g' deploy/operator.yaml
+podman push --authfile "${REGISTRY_AUTH}" "${BUILD_IMAGE:-quay.io/openstack-k8s-operators/neutron-operator:latest}"
```
stuggi, line 9: ah sorry, I think my comment was not clear. what I meant is we build a specific version (maybe this should be a required input) and push the version, but also tag the new version as latest with an additional tag. so that we could just use the latest in our dev automation

stuggi: like in https://docs.quay.io/guides/tag-operations.html

build_image.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+oc create -f deploy/crds/neutron_v1_neutronovsagent_crd.yaml
+oc create -f deploy/crds/neutron_v1_neutronsriovagent_crd.yaml
+oc create -f deploy/crds/neutron.openstack.org_ovsnodeosps_crd.yaml
+oc create -f deploy/crds/neutron.openstack.org_ovncontrollers_crd.yaml
+make
+operator-sdk build --image-builder buildah "${BUILD_IMAGE}:${TAG}"
+sed -i 's|REPLACE_IMAGE|"${BUILD_IMAGE}:${TAG}' deploy/operator.yaml
+podman push --authfile "${REGISTRY_AUTH}" "${BUILD_IMAGE}:${TAG}"
```
stuggi, line 9: should this script fail early if all those variables are not set?

PR #22, opened 2020-07-30
Remove ovsagent

mrKisaoLamb:
Default and only support mech-driver for neutron will be ovn,
so we are not need ovs-agent anymore


dprince:
This looks fine. Removing a CRD like this will likely be a little more work if we had users and wanted to upgrade them and have any existing CR's removed automatically. For now though its good. Merge when ready

PR #23, opened 2020-08-06
Migrate operator-sdk version to v0.19.x

mrKisaoLamb:


Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+COPY controllers/ controllers/
+COPY pkg/ pkg/
+COPY templates/ templates/
+
+# Build
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
+
+# Use distroless as minimal base image to package the manager binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/static:nonroot
```
dprince, line 24: I don't think we can use gcr.io/distroless. I actually switch the mariadb-operator back to something similar like we had before: https://github.com/openstack-k8s-operators/mariadb-operator/blob/master/Dockerfile

The main reason is the scripts we use to create the CSV files depend on a few extra binaries.

Also you'll need the csv-generator here.

controllers/ovncontroller_controller.go
```diff
@@ -302,3 +264,11 @@ func newDaemonset(cr *neutronv1.OVNController, cmName string, templatesConfigHas
 
 	return &daemonSet
 }
+
+func (r *OVNControllerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&neutronv1beta1.OVNController{}).
+		Owns(&corev1.Service{}).
+		Owns(&corev1.ConfigMap{}).
```
dprince, line 272: Won't we want to watch the Daemonset too here? I don't see that in the existing code but perhaps adding it is good?

dprince:
I also see you are changing the API to v1beta1 (from v1). You'll need to update the csv-generator for this change or OLM integration will break as this gets validated when the bundle is created (CRD's have to match what is in the CSV).

Dockerfile
mrKisaoLamb: Updated thanks!

controllers/ovncontroller_controller.go
mrKisaoLamb: Done thanks

controllers/neutronsriovagent_controller.go
```diff
@@ -1,163 +1,116 @@
...
-func (r *ReconcileNeutronSriovAgent) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+// Reconcile reconcile keystone API requests
+// +kubebuilder:rbac:groups=neutron.openstack.org,resources=neutronsriovagents,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=neutron.openstack.org,resources=neutronsriovagents/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;create;update;delete;
+// +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;create;update;delete;
+// +kubebuilder:rbac:groups=core,resources=services,verbs=get;list;create;update;delete;
+// +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;create;update;delete;
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;create;update;delete;
```
stuggi, line 61: you need to add the func comment direct above func line in line 62 and have an empty line between the RBAC definition and the comment. otherwise the parser won't recognize the RBAC when you do make manifests

controllers/ovncontroller_controller.go
```diff
@@ -1,124 +1,90 @@
...
-func (r *ReconcileOVNController) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+// Reconcile reconcile keystone API requests
+// +kubebuilder:rbac:groups=neutron.openstack.org,resources=ovncontrollers,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=neutron.openstack.org,resources=ovncontrollers/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;create;update;delete;
+// +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;create;update;delete;
+// +kubebuilder:rbac:groups=core,resources=services,verbs=get;list;create;update;delete;
+// +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;create;update;delete;
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;create;update;delete;
```
stuggi, line 55: same here

controllers/ovsnodeosp_controller.go
```diff
@@ -1,125 +1,90 @@
...
-func (r *ReconcileOVSNodeOsp) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+// Reconcile reconcile keystone API requests
+// +kubebuilder:rbac:groups=neutron.openstack.org,resources=ovsnodeosps,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=neutron.openstack.org,resources=ovsnodeosps/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;create;update;delete;
+// +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;create;update;delete;
+// +kubebuilder:rbac:groups=core,resources=services,verbs=get;list;create;update;delete;
+// +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;create;update;delete;
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;create;update;delete;
```
stuggi, line 55: same here

PROJECT
```diff
@@ -0,0 +1,16 @@
+domain: openstack.org
+layout: go.kubebuilder.io/v2
+repo: github.com/openstack-k8s-operators/neutron-operator
+resources:
+- group: neutron
+  kind: OVSNodeOsp
+  version: v1beta1
```
stuggi, line 7: this requires a change to the compute-node-operator bindata. I did the same to nova-operator as it reflects better the status of the CRD

yazug:
Dockerfile.rhel is re-worked on PR #26

controllers/neutronsriovagent_controller.go
```diff
@@ -1,163 +1,116 @@
...
+func (r *NeutronSriovAgentReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	_ = context.Background()
+	reqLogger := r.Log.WithValues("neutronsriovagent", req.NamespacedName)
 	reqLogger.Info("Reconciling NeutronSriovAgent")
```
stuggi, line 65: we could also just replace all `reqLogger.Info()` with `r.Log.Info()` for the log statements and keep the default `_ = r.Log.WithValues("neutronsriovagent", req.NamespacedName)`, but this is ok

controllers/neutronsriovagent_controller.go
```diff
@@ -1,163 +1,116 @@
...
 	}
 
-	if err := controllerutil.SetControllerReference(instance, commonConfigMap, r.scheme); err != nil {
-		return reconcile.Result{}, err
+	if err := controllerutil.SetControllerReference(instance, commonConfigMap, r.Scheme); err != nil {
+		return ctrl.Result{}, err
 	}
 
 	// Create additional host entries added to the /etc/hosts file of the containers
 	ospHostAliases, err = util.CreateOspHostsEntries(commonConfigMap)
```
stuggi, line 95: we no longer need to add the hosts entries as we now have dns service from OCP. maybe in future we want to add it again where external services are required.

controllers/neutronsriovagent_controller.go
```diff
@@ -436,3 +389,10 @@ func newDaemonset(cr *neutronv1.NeutronSriovAgent, cmName string, configHash str
 
 	return &daemonSet
 }
+
+// SetupWithManager x
+func (r *NeutronSriovAgentReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&neutronv1beta1.NeutronSriovAgent{}).
```
stuggi, line 396: we'd need to add here the objects to watch
```
		Owns(&corev1.ConfigMap{}).
		Owns(&appsv1.DaemonSet{}).
```

controllers/ovncontroller_controller.go
```diff
@@ -1,124 +1,90 @@
...
+func (r *OVNControllerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	_ = context.Background()
+	reqLogger := r.Log.WithValues("ovncontroller", req.NamespacedName)
 	reqLogger.Info("Reconciling OVNController")
```
stuggi, line 59: same as above, we could just use `r.Log.Info()`

stuggi (commented):
the sriov agent needs to watch the resources it creates (CM, daemonset) and we don't have to add the hosts entries atm as we remove the "old" common-config. I have a patch for that for review 
 for the nova-operator to see how we can do this https://github.com/openstack-k8s-operators/nova-operator/pull/42 . some parts should then be moved to lib-common, so no need todo this in this patch :)

luis5tb:
We can remove the things related to configuring ml2/ovs (like ovs agent and things like that), but not the ovs itself. That is still needed, and it is shared between OVN from OSP and OVN from OCP

dprince (approved):

luis5tb:
other than the versioning thing and the missing readme file, looks good to me

PR #24, opened 2020-08-24
Initial support for ovn-multibridge at neutron operator

luis5tb:


deploy/crds/neutron.openstack.org_v1_ovsnodeosp_cr.yaml
```diff
@@ -7,6 +7,6 @@ spec:
   ovsNodeOspImage: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:7d6ba1bcc4f403733e44cd2a008dafe1887501f7a3c74084566a4f4d1e46250d
   serviceAccount: neutron
   roleName: worker-osp
-  ovnSbRemote: 192.168.25.100
+  ovnSbRemote: tcp:172.30.71.55:6642,tcp:172.30.110.147:6642,tcp:172.30.51.24:6642
```
mdbooth, line 10: Note that the pattern we've established with these in dev-tools is that operators are deployed by OLM and the CRs are copied into, and customised in, the dev-tools repo. So these changes would also need to end up there when we land this.

templates/ovncontroller/ovn.sh
```diff
@@ -13,6 +13,9 @@ else
     # ovn-appctl is not present. Use openvswitch run dir path.
     OVNCTL_DIR=openvswitch
 fi
-exec ovn-controller unix:/var/run/openvswitch/db.sock -vfile:off \
+
+HOSTNAME=`/usr/bin/hostname 2>/dev/null`
```
mdbooth, line 17: FWIW the 2>/dev/null is redundant here as it won't write anything to stderr, and if it did that wouldn't end up in the HOSTNAME variable. A more concise version would be:

HOSTNAME=$(hostname)

You could also pass it from the pod with the downward api:

      env:
        - name: HOSTNAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name

https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#the-downward-api

mdbooth (approved):
LGTM. I'll work on generating an ConfigMap to pull ovnSbRemote from that an operator might generate.

Some minor comments.

mrKisaoLamb (approved):
lgtm

deploy/crds/neutron.openstack.org_ovsnodeosps_crd.yaml
```diff
@@ -34,9 +34,6 @@ spec:
             nic:
               description: NIC for ovn encap ip
               type: string
-            ovnSbRemote:
-              description: ovn south db ip
-              type: string
```
mdbooth, line 39: Should we add the name of the ovn-connection configmap to the CRD? Don't do this because I'm not sure, and this is fine for now.

mdbooth (approved):
Nice improvement.

PR #26, opened 2020-08-27
Rework Dockerfile and add Dockerfile.rhel

yazug:
- use ARG's for variants between the 2 files
- Dockerfile.rhel is intended to build in
  OSBS using cachito external sources bundle
- dropped CGO_ARCH and target from building
- copy all of the source in and then setup
  for copying into final image
- go builds should be statically linked now
- crds are copied into a bundle folder first
  and then copied as a whole into final image
- Dockerfile should continue to work as
  expected for local builds
- Dockerfile.rhel should work for building downstream

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,50 +1,64 @@
-# Build the manager binary
-FROM golang:1.13 as builder
+# golang-builder is used in OSBS build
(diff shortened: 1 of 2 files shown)


yazug:
dependent on PR #23

PR #33, opened 2020-10-01
Add envtests for Ovn controller and Ovs node

mrKisaoLamb:
Add script to download kubebuilder and run test
Now to run tests from controller we can use simply make tests


stuggi:
ooold

PR #38, opened 2022-08-01
Modernize Neutron Operator

karelyatin:


config/rbac/auth_proxy_service.yaml
```diff
@@ -4,11 +4,12 @@ metadata:
   labels:
     control-plane: controller-manager
   name: controller-manager-metrics-service
-  namespace: system
+  namespace: openstack
```
abays, line 7: We've kept this as `system` in the other operators.  Was there a specific reason it became `openstack` here?

go.mod
```diff
@@ -1,23 +1,90 @@
...
-	github.com/prometheus/common v0.7.0
-	golang.org/x/net v0.0.0-20200822124328-c89045814202 // indirect
-	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
-	k8s.io/api v0.18.6
-	k8s.io/apimachinery v0.18.6
-	k8s.io/client-go v0.18.6
-	sigs.k8s.io/controller-runtime v0.6.2
+	github.com/openstack-k8s-operators/keystone-operator v0.0.0-20220727152756-f13e087e4199
+	github.com/openstack-k8s-operators/keystone-operator/api v0.0.0-20220727152756-f13e087e4199
+	github.com/openstack-k8s-operators/lib-common v0.0.0-20220727151739-0ac9fed768b3
```
abays, line 12: At the end of last week, we finished reorganizing the module/package layout of `lib-common`.  We'll either want to use that new version here or in a follow-up to this PR.

config/rbac/auth_proxy_service.yaml
karelyatin: @abays No no reason to keep openstack here, it's just miss, /me will update.

go.mod
karelyatin: Both option works for me.

pkg/neutronapi/dbsync.go
```diff
@@ -47,7 +49,7 @@ func DbSyncJob(cr *neutronv1beta1.NeutronAPI, scheme *runtime.Scheme) *batchv1.J
 								},
 								{
 									Name:  "DatabaseHost",
-									Value: cr.Spec.DatabaseHostname,
+									Value: cr.Spec.DatabaseInstance,
 								},
```
abays, line 53: The `DatabaseX` env vars shouldn't be needed here, as they are given to the init container and used there instead.

pkg/neutronapi/dbsync.go
```diff
@@ -73,15 +75,13 @@ func DbSyncJob(cr *neutronv1beta1.NeutronAPI, scheme *runtime.Scheme) *batchv1.J
 			},
 		},
 	}
-	initContainerDetails := common.InitContainer{
+	initContainerDetails := InitContainer{
 		ContainerImage: cr.Spec.ContainerImage,
-		DatabaseHost:   cr.Spec.DatabaseHostname,
+		DatabaseHost:   cr.Spec.DatabaseInstance,
```
abays, line 80: You will want to use `cr.Status.DatabaseHostname` as that should be have been populated with the proper hostname during your `reconcileInit`: https://github.com/openstack-k8s-operators/neutron-operator/blob/379c6f50180863180b745d6f4672701f35c85dc3/controllers/neutronapi_controller.go#L241

pkg/neutronapi/dbsync.go
```diff
@@ -73,15 +75,13 @@ func DbSyncJob(cr *neutronv1beta1.NeutronAPI, scheme *runtime.Scheme) *batchv1.J
...
 		Database:       Database,
-		NeutronSecret:  cr.Spec.NeutronSecret,
-		NovaSecret:     cr.Spec.NovaSecret,
+		NeutronSecret:  cr.Spec.Secret,
```
abays, line 82: Another approach we have been using for secrets is to use `XPasswordSelector` in the CR spec so that users can specify which keys in the secret contain the values needed for certain considerations: https://github.com/openstack-k8s-operators/keystone-operator/blob/038a3f1b65144cc53313cfb3a9952645a79f7a37/pkg/keystone/dbsync.go#L90-L92.  You then adjust the init container env var code to uses these selectors instead.

pkg/neutronapi/deployment.go
```diff
@@ -0,0 +1,148 @@
...
+		// https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
+		//
+		livenessProbe.HTTPGet = &corev1.HTTPGetAction{
+			Path: "/",
+			Port: intstr.IntOrString{Type: intstr.Int, IntVal: int32(NeutronPublicPort)},
+		}
+		readinessProbe.HTTPGet = &corev1.HTTPGetAction{
+			Path: "/",
+			Port: intstr.IntOrString{Type: intstr.Int, IntVal: int32(NeutronPublicPort)},
+		}
```
abays, line 77: One thing to check here: do these `/` queries return a 200 response code?  If not, they will generate warnings every few seconds.  We had to change to a different path for Glance, for instance: https://github.com/openstack-k8s-operators/glance-operator/blob/0df23ef3490a6e0ab83f66fa355174d2e11bef0b/pkg/glance/deployment.go#L73-L80

templates/neutronapi/config/neutron-api-config.json
```diff
@@ -1,12 +1,18 @@
 {
-  "command": "/usr/bin/neutron-server --config-file /usr/share/neutron/neutron-dist.conf --config-file /etc/neutron/neutron.conf",
+  "command": "/usr/bin/neutron-server --config-file /usr/share/neutron/neutron-dist.conf --config-file /etc/neutron/neutron.conf --config-dir /etc/neutron/neutron.conf.d",
```
abays, line 2: Does the combination of `--config-file` and `--config-dir` work appropriately?  I know that for Glance, for instance, you can't use them together (one ends up being ignored).  I need a quick search for Neutron docs in this regard, but couldn't find anything.

controllers/neutronapi_controller.go
```diff
@@ -89,327 +98,467 @@ func (r *NeutronAPIReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
 	}
-	return nil
+	configMapVars[ospSecret.Name] = common.EnvValue(hash)
 
-}
+	// check for required ovn-connection configMap
+	ovnConnection, hash, err := common.GetConfigMapAndHashWithName(ctx, helper, instance.Spec.OVNConnectionConfigMap, instance.Namespace)
+	if err != nil {
+		return ctrl.Result{RequeueAfter: time.Second * 10}, err
+	}
```
abays, line 386: In the CRD definition, `OVNConnectionConfigMap` is not flagged as required.  Should this only check for the config map if `OVNConnectionConfigMap` is not the empty string?  Or should the CRD be changed so that this indeed required?

pkg/neutronapi/deployment.go
karelyatin: Yes it returned 200 response so i used that, but yes can check if some better path can be used instead.

templates/neutronapi/config/neutron-api-config.json
karelyatin: From my tests using both together worked fine, from functionality wise i think it should be same for both glance and neutron as both using oslo.config.

pkg/neutronapi/dbsync.go
karelyatin: Sure will check and fix

pkg/neutronapi/dbsync.go
karelyatin: Hmm i have used it in some places, will check and generalize to other places too.

templates/neutronapi/config/neutron.conf
```diff
@@ -28,18 +28,19 @@ ovn_sb_connection = {{ .SBConnection }}
...
+auth_url = {{ .KeystonePublicURL }}
 # XXX(mdbooth): Add memcached
 #memcached_servers = controller:11211
 auth_type = password
 project_domain_name = Default
 user_domain_name = Default
 project_name = service
-username = neutron
+username = {{ .ServiceUser }}
+password = foobar123
```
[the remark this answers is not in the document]
karelyatin: Yes right, will cleanup extras

templates/neutronapi/config/neutron-api-config.json
abays: Okay, just wanted to check.  With regards to Glance, I'll just point this out:
```
If --config-dir is set, then --config-file is ignored.
```
That's from this doc: https://docs.openstack.org/glance/latest/configuration/configuring.html

controllers/neutronapi_controller.go
karelyatin: For functionality wise it's needed, so i think would make it as required would be better.
But considering the Data in this config map comes from ovn Resources, so when ovn-operator will be ported, these can be set in OVN resource Status and those can be used instead of using config map.

Currently i used https://github.com/openstack-k8s-operators/osp-director-dev-tools/tree/master/ansible/roles/cnosp/files/ovn for setting up ovn resources and create OVNConnectionConfigMap based on the setup.

abays: Understood.  I see you have a PR for Neutron for `install_yamls` as well.  It sounds like the `OVN` automation from `osp-director-dev-tools` might need to be added to `install_yamls` as well.

controllers/neutronapi_controller.go
```diff
@@ -89,327 +98,467 @@ func (r *NeutronAPIReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
+	authURL, err := keystoneAPI.GetEndpoint(common.EndpointPublic)
+	if err != nil {
+		return err
+	}
+	templateParameters := make(map[string]interface{})
+	templateParameters["ServiceUser"] = instance.Spec.ServiceUser
+	templateParameters["KeystonePublicURL"] = authURL
 
+	ovnConnection, _, err := common.GetConfigMapAndHashWithName(ctx, h, instance.Spec.OVNConnectionConfigMap, instance.Namespace)
```
abays, line 500: I think you're already getting this in `reconcileNormal`, so you could pass it to this function to avoid querying `etcd` again.

controllers/neutronapi_controller.go
```diff
@@ -89,327 +98,467 @@ func (r *NeutronAPIReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
+			Labels:        cmLabels,
+			ConfigOptions: templateParameters,
+		},
+		// CustomConfigMap
+		{
+			Name:      fmt.Sprintf("%s-config-data-custom", instance.Name),
+			Namespace: instance.Namespace,
+			Type:      common.TemplateTypeCustom,
+			Labels:    cmLabels,
+		},
```
abays, line 533: I'm not sure you need this.  You're using the same pattern we have in the other modernized operators, where `custom.conf` is generated from the `CustomServiceConfig` CRD parameter.  In that case, the `custom.conf` is part of the `<instance.Name>-config-data` config map, so you don't need to mount an extra config map.  See https://github.com/openstack-k8s-operators/glance-operator for reference, as that uses the same pattern to handle custom config without requiring an extra config map for that custom config.

abays: And if you don't need it here, you can also remove all code elsewhere that references `<instance.Name>-config-data-custom`.

karelyatin: Yes right, it's extra, cleaned all it's references.

go.mod
abays: Okay, one of us can follow-up after we finish with this PR

pkg/neutronapi/dbsync.go
```diff
@@ -1,28 +1,34 @@
...
+) *batchv1.Job {
 
 	runAsUser := int64(0)
 	initVolumeMounts := GetInitVolumeMounts()
 	volumeMounts := GetAPIVolumeMounts()
 	volumes := GetAPIVolumes(cr.Name)
+	cmLabels := common.GetLabels(cr, common.GetGroupLabel(ServiceName), map[string]string{})
+
+	envVars := map[string]common.EnvSetter{}
+	envVars["KOLLA_CONFIG_FILE"] = common.EnvValue("/var/lib/config-data/merged/db-sync-config.json")
```
abays, line 24: Optional: could make the `/var/lib/config-data/merged/db-sync-config.json` a const in your `const.go` in the package

.github/workflows/build-neutron-operator.yaml
```diff
@@ -0,0 +1,200 @@
...
+    needs: [ check-secrets, build-neutron-operator ]
+    name: neutron-operator-bundle
+    runs-on: ubuntu-latest
+    if: needs.check-secrets.outputs.have-secrets == 'true'
+
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.18.0
```
stuggi, line 85: should be 1.18.x

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+jobs:
+  test:
+    name: github (govet, golint and gotest)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.18.0
```
stuggi, line 13: should be 1.18.x

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,42 @@
...
+        run: ./openstack-k8s-operators-ci/test-runner/gotest.sh
+
+  golangci:
+    name: github (golangci)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.18.0
```
stuggi, line 35: should be 1.18.x

Makefile
```diff
@@ -23,96 +58,213 @@ else
...
-generate: controller-gen
-	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+## Tool Binaries
+KUSTOMIZE ?= $(LOCALBIN)/kustomize
+CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
+ENVTEST ?= $(LOCALBIN)/setup-envtest
+
+## Tool Versions
+KUSTOMIZE_VERSION ?= v3.8.7
+CONTROLLER_TOOLS_VERSION ?= v0.9.0
```
stuggi, line 164: we introduce sub modules for the api itself which can then be used by other operator without getting all the dependencies. We have this now in the keystone operator as a first one and I had issues with 0.9.0 which works fine with 0.9.2, therefore might be good even when we not yet have the api sub module in here to use it.

Makefile
```diff
@@ -23,96 +58,213 @@ else
...
+
+KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
+.PHONY: kustomize
+kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
+$(KUSTOMIZE): $(LOCALBIN)
+	curl -s $(KUSTOMIZE_INSTALL_SCRIPT) | bash -s -- $(subst v,,$(KUSTOMIZE_VERSION)) $(LOCALBIN)
```
stuggi, line 170: should have a `rm -f $(LOCALBIN)/kustomize || true` before the curl, otherwise subsequent runs will stop with a need to be removed massage or so. thats something we need to add to other operators as well.

config/manifests/bases/neutron-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,48 @@
...
+    type: MultiNamespace
+  - supported: true
+    type: AllNamespaces
+  keywords:
+  - OpenStack
+  - Network
+  - Neutron
+  links:
+  - name: Neutron Operator
+    url: https://neutron-operator.domain
```
stuggi, line 43: should be https://github.com/openstack-k8s-operators/neutron-operator

stuggi (commented):
just some minor prio comments, other then that I think we are good to get this landed as the initial version and go from there. Thanks!

Makefile
karelyatin: Updated

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,16 @@
...
+
+configMapGenerator:
+- files:
+  - controller_manager_config.yaml
+  name: manager-config
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: quay.io/ykarel/neutron-operator
```
stuggi, line 15: don't think you wanted to add this from your latest pushes?

karelyatin: yes right, my bad, updating

stuggi (approved):
/lgtm to get in as the initial version and follow up with additional PRs on what Andrew commented.

abays (approved):
/lgtm

PR #42, opened 2022-09-27
Update to use latest dependencies with new KeystoneEndpoint

stuggi:
The keystone-operator introduced a new CRD KeystoneEndpoint to register endpoints separat from the service and service user. 
This is required by some services.


karelyatin (approved):
LGTM, also works fine in my local tests.

abays (approved):
/lgtm

PR #46, opened 2022-10-31
Use OVNDBCluster Status object for NB/SB info

karelyatin:
NB/SB DB Connection info's are available in OVNDBCluster Status,
let's use it from it instead of ovn-connection config map.


controllers/neutronapi_controller.go
```diff
@@ -655,12 +635,16 @@ func (r *NeutronAPIReconciler) generateServiceConfigMaps(
 	ctx context.Context,
 	h *helper.Helper,
 	instance *neutronv1beta1.NeutronAPI,
-	ovnConnection *corev1.ConfigMap,
 	envVars *map[string]env.Setter,
 ) error {
 	// Create/update configmaps from templates
 	cmLabels := labels.GetLabels(instance, labels.GetGroupLabel(neutronapi.ServiceName), map[string]string{})
 
+	dbmap, err := ovndbclient.GetDBEndpoints(ctx, h, instance.Namespace, map[string]string{})
```
abays, line 643: I think I understand the use of the `ovndbclient` here -- to import the code and avoid duplicating certain logic across this operator and others that might need to acquire such things as you're doing here for the DB endpoints.  The unfortunate downside to this approach though is that now the Neutron operator inherits all of the OVN operator dependencies.  So I think there are two other options here:

1. Accept that code might be duplicated across operators and put the logic for `GetDBEndpoints` in this operator under a `pkg` package.
2. Move the `GetDBEndpoints` logic in the OVN operator into its respective `api` module and import that smaller module instead.  I'm aware that we're at best ambivalent about the "move-to-api-module" approach, but I just thought I'd mention it.  Anyhow, an example of where we have done this elsewhere is the Keystone operator, where we removed `pkg/external` and moved the code into `api` to allow operators to only import that module as opposed to the whole Keystone operator: https://github.com/openstack-k8s-operators/keystone-operator/commit/155ef0bbd106500495c980dd0faea3559d3cbcb1

karelyatin: Thanks Andrew, will looking into creating api module in ovn-operator.

karelyatin: Done

abays (approved):
/lgtm

PR #50, opened 2022-11-16
Fix rbac for OVNDBCluster

karelyatin:
Previous commit wrongly set it to ovndbcluster,
it should be "ovndbclusters" instead.


karelyatin:
Self approving, typo from previous commit.

PR #64, opened 2022-11-30
Update initial delay for probes

karelyatin:
Neutron server sometimes takes time to start
as multiple workers needs to be started and also
slow infra can further delay it.

After 3(Default failureThreshold) liveness probe failures pod get's restarted and can stuck into restart loop. With this patch increasing initial delay to 20s for both liveness and readyness probes and
also increase periodseconds to 5s in liveness probe.

Not removing TODO comment from the probes as we would tune these further later.

### DIFF
--- a/pkg/neutronapi/deployment.go
+++ b/pkg/neutronapi/deployment.go
@@ -40,14 +40,14 @@ func Deployment(
 	livenessProbe := &corev1.Probe{
 		// TODO might need tuning
 		TimeoutSeconds:      5,
-		PeriodSeconds:       3,
-		InitialDelaySeconds: 3,
+		PeriodSeconds:       5,
+		InitialDelaySeconds: 20,
 	}
 	readinessProbe := &corev1.Probe{
 		// TODO might need tuning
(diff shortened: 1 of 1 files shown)


pkg/neutronapi/deployment.go
```diff
@@ -40,14 +40,14 @@ func Deployment(
 	livenessProbe := &corev1.Probe{
 		// TODO might need tuning
```
slawqo, line 41: maybe You can remove that TODO now?

pkg/neutronapi/deployment.go
```diff
@@ -40,14 +40,14 @@ func Deployment(
...
 		TimeoutSeconds:      5,
-		PeriodSeconds:       3,
-		InitialDelaySeconds: 3,
+		PeriodSeconds:       5,
+		InitialDelaySeconds: 20,
 	}
 	readinessProbe := &corev1.Probe{
 		// TODO might need tuning
```
slawqo, line 47: ditto

slawqo (approved):

PR #66, opened 2022-12-04
add_service_db_finalizer_logic

pinikomarov:
Ever since  : https://github.com/openstack-k8s-operators/lib-common/pull/87

Podified operators which use service dbs and the db.CreateOrPatchDB method ,
now will get the finalizer set on the db automatically when using the newest lib-common,
but they won't have the code remove it on cleanup i.e reconcileDelete,
which might make the delete operator action hang.

This adds the service Db removal code in the operator delete (reconcileDelete)
function to work with the above.

Also updated lib-common to latest as to not break lib-common's exported functions


weinimo:
Are you saying that all operators that use db.CreateOrPatchDB() should do the same?

pinikomarov:
> Are you saying that all operators that use db.CreateOrPatchDB() should do the same?

yes , otherwise the func reconcileDelete will miss out on removing all the finalizers created during the reconcile etc. And deleting that service db might hang.
Hi @stuggi if there is an exception here and db.DeleteFinalizer should not be used , please let me us know.

karelyatin:
@pinikomarov are more changes pending on this? just trying to understand why PR in Draft state. Change in itself LGTM.

controllers/neutronapi_controller.go
```diff
@@ -203,6 +203,18 @@ func (r *NeutronAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	// remove db finalizer first
+	db, err := database.GetDatabaseByName(ctx, helper, instance.Name)
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return ctrl.Result{}, err
+	}
+
+	if !k8s_errors.IsNotFound(err) {
+		if err := db.DeleteFinalizer(ctx, helper); err != nil {
+			return ctrl.Result{}, err
+		}
```
karelyatin, line 215: Would be good to add a log like KestoneEndpoint and Service.
util.LogForObject(helper, "Removed finalizer from our MariaDBDatabase", instance)

pinikomarov: Sure , done

karelyatin (approved):
/lgtm

PR #91, opened 2023-01-17
Use internal endpoint to interact with Keystone/Nova

kajinamit:
Internal endpoint is meant for interaction between services. This ensures neutron uses internal endpoints when it sends requests to the other services.


beagles (approved):
lgtm

PR #94, opened 2023-01-18
Allow attaching service pods to additional networks

stuggi:
For network isolation outgoing traffic, service pods need to be attached additional networks. To achive this, the the `k8s.v1.cni.cncf.io/networks` annotation needs to be set on the pods. This adds a networkAttachmentDefinitions parameter to the CRD where it can be specified to which additional networks the service pods should be attached to.

Depends-On: openstack-k8s-operators/lib-common#133


abays (approved):
/lgtm

PR #107, opened 2023-02-02
Fix pattern to exclude bin

kajinamit:
Current pattern can match subdirectories and makes some files in templates directories excluded.

Signed-off-by: Takashi Kajinami <tkajinam@redhat.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@
 *.dll
 *.so
 *.dylib
-bin
+/bin
 testbin/*
 
 # Test binary, build with `go test -c`


.gitignore
```diff
@@ -5,7 +5,7 @@
 *.dll
 *.so
 *.dylib
-bin
+/bin
```
karelyatin, line 8: seems bin/* would be also be ok for the case you mentioned, i hit it previously with ovn-operator and kept it similar pattern as testbin/*.

kajinamit: Yeah. That also works.

The reason I used `/bin` was it is used by a few other operators such as cinder/glance/nova and tried to make this consistent across operators.

karelyatin: Ok Thanks, that works for me.

karelyatin (approved):

PR #108, opened 2023-02-03
Support exposing service via metallb

stuggi:
* renames NetworkAttachmentDefinitions to NetworkAttachments
* introduce externalEndpoints Parameters to configure metallb service using the pre-created metallb IPAddressPool and L2Advertisement . 

E.g. to expose the api on an internalapi network which
* should be registered as the internal endpoint
* its metallb ipaddresspool is configured in an osp-internalapi object
* and request a specific loadbalancerIP which is shared with other services (list of IPs for multi stack ipv4/ipv6)

```
...
spec:
...
      externalEndpoints:
      - endpoint: internal
        loadbalancerIPs:
        - 172.17.0.202
        ipAddressPool: osp-internalapi
        sharedIP: true
```

* removes registration of admin endpoint

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/160


abays (approved):
/lgtm

PR #113, opened 2023-02-09
Use time.Duration for timeout

kajinamit:
instead of int. This allows us to use more fine grained timeout value such as 0.1, and encodes the unit in the type itself.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/155
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/166


abays (approved):
/lgtm

Linting failure looks like a false positive

gibizer:
> /lgtm
> 
> Linting failure looks like a false positive

Nope. It fails as go mod tidy isn't run after the lib-common version bump. I'm fixing it...

gibizer:
> > /lgtm
> > Linting failure looks like a false positive
> 
> Nope. It fails as go mod tidy isn't run after the lib-common version bump. I'm fixing it...

Done

gibizer (approved):

abays (approved):
/lgtm

PR #117, opened 2023-02-10
Update operator-lint to latest

amolkahat:
This will fix the pre-commit failures.

openshift/release: [36102](https://github.com/openshift/release/pull/36012)

Signed-off-by: Amol Kahat <amolkahat@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -302,5 +302,5 @@ gowork: ## Generate go.work file to support our multi module repository
 
 .PHONY: operator-lint
 operator-lint: gowork ## Runs operator-lint
-	GOBIN=$(LOCALBIN) go install github.com/gibizer/operator-lint@v0.1.0
+	GOBIN=$(LOCALBIN) go install github.com/gibizer/operator-lint@latest
 	go vet -vettool=$(LOCALBIN)/operator-lint ./... ./api/...
--- a/api/bases/neutron.openstack.org_neutronapis.yaml
+++ b/api/bases/neutron.openstack.org_neutronapis.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.10.0
+    controller-gen.kubebuilder.io/version: v0.9.2
   creationTimestamp: null
   name: neutronapis.neutron.openstack.org
 spec:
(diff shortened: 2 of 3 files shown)


config/crd/bases/neutron.openstack.org_neutronapis.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.10.0
+    controller-gen.kubebuilder.io/version: v0.9.2
```
sandeepyadav93, line 6: Just curious, what's the reason behind this?

amolkahat: This got updated after pre-commit run.

karelyatin: can you try again after dropping rm bin/controller-gen from local neutron-operator clone. With that these files will not change, v0.10.0 is the current version as per CONTROLLER_TOOLS_VERSION ?= v0.10.0 in Makefile

api/bases/neutron.openstack.org_neutronapis.yaml
```diff
@@ -113,7 +113,7 @@ spec:
                   database: NeutronDatabasePassword
                   service: NeutronPassword
                 description: PasswordSelectors - Selectors to identify the DB and
-                  ServiceUser and NoveService User password from the Secret
+                  ServiceUser password from the Secret
```
karelyatin, line 116: same here

Makefile
```diff
@@ -302,5 +302,5 @@ gowork: ## Generate go.work file to support our multi module repository
 
 .PHONY: operator-lint
 operator-lint: gowork ## Runs operator-lint
-	GOBIN=$(LOCALBIN) go install github.com/gibizer/operator-lint@v0.1.0
+	GOBIN=$(LOCALBIN) go install github.com/gibizer/operator-lint@latest
```
karelyatin, line 305: from past discussion i recall intentions were to use pinned version of operator-lint so projects can move to more recent version whenver it suits to project requirement. or the decision changed or i missed something?

can you also share the Error which this change is fixing to understand better.

karelyatin (changes requested):

config/crd/bases/neutron.openstack.org_neutronapis.yaml
amolkahat: Well this works.

Makefile
amolkahat: Well fixing some erros with pre-commit.. This is not relevant patch to the errors. Closing.

amolkahat:
Closing. not relevant.

PR #118, opened 2023-02-13
Replace deprecated functions

lewisdenny:
Replace the following deprecated functions:

db.WaitForDBCreated -> db.WaitForDBCreatedWithTimeout 
db.CreateOrPatchDB -> db.CreateOrPatchDBByName

The functionality stays the same.


controllers/neutronapi_controller.go
```diff
@@ -272,9 +272,10 @@ func (r *NeutronAPIReconciler) reconcileInit(
 		},
 	)
 	// create or patch the DB
-	ctrlResult, err := db.CreateOrPatchDB(
+	ctrlResult, err := db.CreateOrPatchDBByName(
 		ctx,
 		helper,
+		"openstack",
```
bshephar, line 278: This should use a variable to allow users to choose the DB they're updating. Seems as though Neutron operator is using `DatabaseInstance` for this:
https://github.com/openstack-k8s-operators/neutron-operator/blob/master/api/v1beta1/neutronapi_types.go#L52-L56
https://github.com/openstack-k8s-operators/neutron-operator/blob/master/config/samples/neutron_v1beta1_neutronapi.yaml#L11


Note the implementation of this in the nova-operator:
https://github.com/openstack-k8s-operators/nova-operator/blob/master/controllers/nova_controller.go#L512-L516

bshephar (changes requested):
Small adjustment required here.

controllers/neutronapi_controller.go
lewisdenny: Thanks @bshephar, 

Sadly I haven't been able to test this change in my dev environment but I can see `instance.Spec.DatabaseInstance` being used just above on line 271 and from the code side I see no issues.

bshephar (approved):

controllers/neutronapi_controller.go
karelyatin: Need to fix tests as per this change as we using different value of database Instance in test(https://github.com/openstack-k8s-operators/neutron-operator/blob/master/test/functional/base_test.go#L78)
So basically need to fix[1][2] with "openstack" -> GetNeutronAPI(neutronAPIName).Spec.DatabaseInstance and [3] with "hostname-for-openstack" -> "hostname-for-" + NeutronAPI.Spec.DatabaseInstance

[1] https://github.com/openstack-k8s-operators/neutron-operator/blob/master/test/functional/neutronapi_controller_test.go#L329
[2] https://github.com/openstack-k8s-operators/neutron-operator/blob/master/test/functional/neutronapi_controller_test.go#L376
[3] https://github.com/openstack-k8s-operators/neutron-operator/blob/master/test/functional/neutronapi_controller_test.go#L343
, GetNeutronAPI(neutronAPIName).Spec.DatabaseInstance,

karelyatin (changes requested):
comment inline

karelyatin (approved):
/lgtm

PR #119, opened 2023-02-13
Add basic KUTTL tests for neutron-operator

cescgina:
These tests cover basic operations: deployment, scale up, scale down and clean up.

The tests use code from the install_yamls repo to reuse some operations like
the deployment of the neutron operator, as well as the cleanup from previous runs.

Test are thought to be run from the `install_yamls` repo, see the following
PR: https://github.com/openstack-k8s-operators/install_yamls/pull/99

There are three ways to run the tests:
1) Run `make neutron_kuttl`, this will only work if you want to run the test
from the master branch.
2) Run `make neutron_kuttl NEUTRON_REPO=ovs-operator._repo_url
NEUTRON_BRANCH=branch_name` this is useful if you want to run the tests from
a fork
3) Run `make neutron_kuttl
NEUTRON_KUTTL_CONF=/path/to/neutron-operator/kuttl-test.yaml
NEUTRON_KUTTL_DIR=/path/to/neutron-operator/tests/kuttl/tests/` this is
useful to run local changes to the tests

### DIFF
--- a/kuttl-test.yaml
+++ b/kuttl-test.yaml
@@ -1,0 +1,25 @@
+#
+# EXECUTION (from install_yamls repo root):
+#
+#   make neutron_kuttl
+#
+# ASSUMPTIONS:
+#
+# 1. Latest version of kuttl is installed at /usr/local/bin/kubectl-kuttl
+#    - wget https://github.com/kudobuilder/kuttl/releases/download/v0.11.1/kubectl-kuttl_0.11.1_linux_x86_64
+#    - mv kubectl-kuttl_0.11.1_linux_x86_64 /usr/local/bin/kubectl-kuttl
+#    - chmod 755 /usr/local/bin/kubectl-kuttl
+# 2. An OCP 4.10+ CRC cluster with Podified Operators has been deployed
+# 3. CLI user has access to $KUBECONFIG
+# 4. The environment variable INSTALL_YAMLS is set to the the path of the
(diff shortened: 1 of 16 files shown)


kuttl-test.yaml
```diff
@@ -0,0 +1,25 @@
...
+# 3. CLI user has access to $KUBECONFIG
+# 4. The environment variable INSTALL_YAMLS is set to the the path of the
+#    install_yamls repo
+
+apiVersion: kuttl.dev/v1beta1
+kind: TestSuite
+reportFormat: JSON
+reportName: kuttl-test-neutron
+namespace: openstack
+timeout: 180
```
marios, line 22: adding link incase useful for other reviewers

these config attributes are described there https://kuttl.dev/docs/testing/reference.html#testsuite
e.g. timeout is in seconds

marios (commented):
lgtm didn't spot something

kuttl-test.yaml
marios: though perhaps the subject of the pull request can be updated

it is the same as the install_yamls one https://github.com/openstack-k8s-operators/install_yamls/pull/99 where it makes more sense

cescgina: Ah, you're right, I've mixed them up, since I created one after the other, I'll update it, thanks

abays (approved):
/lgtm

PR #122, opened 2023-02-16
Drop paths-ignore from build operator workflow

sandeepyadav93:
We build containers and push to registry in post merge workflow. Not building container for each hash can cause issue becuase of missing container tags. Pin-custom-bundle-dockerfile.sh (openstack-operator) can pin to hash for which images are not available.

### DIFF
--- a/.github/workflows/build-neutron-operator.yaml
+++ b/.github/workflows/build-neutron-operator.yaml
@@ -4,20 +4,6 @@ on:
   push:
     branches:
       - '*'
-    paths-ignore:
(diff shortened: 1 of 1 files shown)


karelyatin:
But seems this will build/push images even for changes that doesn't change target image and that looks totally unnecessary. 
Sorry but /me not aware about Pin-custom-bundle-dockerfile.sh and when it get's used. May be that script should check for available latest hash and use that?

dprince (approved):

PR #134, opened 2023-04-06
Add golangci-lint config

gibizer:
* Bumps the timeout from 1min to 5min to allow slower CI runs to pass
* Enabled common set of linter


karelyatin (approved):
/lgtm

PR #138, opened 2023-04-19
Add local templates export to MakeFile

lewisdenny:
This is required when running operator locally and has been added to some other operators already such as [1]

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/master/Makefile#L147

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,6 +128,7 @@ build: generate fmt vet ## Build manager binary.
 
 .PHONY: run
 run: export ENABLE_WEBHOOKS?=false
+run: export OPERATOR_TEMPLATES=./templates/
 run: manifests generate fmt vet ## Run a controller from your host.
 	/bin/bash hack/clean_local_webhook.sh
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -128,6 +128,7 @@ build: generate fmt vet ## Build manager binary.
 
 .PHONY: run
 run: export ENABLE_WEBHOOKS?=false
+run: export OPERATOR_TEMPLATES=./templates/
```
karelyatin, line 131: Let's change it to export OPERATOR_TEMPLATES?=./templates/ so it can be overridden from CLI

karelyatin (changes requested):

Makefile
lewisdenny: Oh nice, TIL: https://www.gnu.org/software/make/manual/make.html#index-conditional-variable-assignment

karelyatin (approved):
/lgtm

PR #142, opened 2023-04-26
Remove unused constant for admin endpoint

kajinamit:
The constant has not been used since we stopped creating admin endpoint.


karelyatin (approved):
/lgtm

PR #143, opened 2023-05-04
Adapt to split test helpers in lib-common

gibizer:
The test module in lib-common has been split in
https://github.com/openstack-k8s-operators/lib-common/pull/252 . The new modules/test-operators module the one that has the full set of helpers we need to use in our tests. The previously used modules/test is reduced to only contain non openstack-k8s-operators specific helpers.

This patch simply change the import to use the new module

PR #144, opened 2023-05-04
Lib common test helper usage

gibizer:
* Add GINKGO_ARGS to `make test` to allow running individual test cases
* Removed a bunch of duplicated helper code as they are now available in lib-common.

Depends-on: #143


karelyatin (approved):
/lgtm

PR #145, opened 2023-05-05
Add GINKGO_ARGS env var to test target

karelyatin:
With this change you can pass arbitrary parameters to the ginkgo test executor via the make target.

E.g. if you want to run only a subset of test sequentially failing on the first error then you can do that with:

make test GINKGO_ARGS="--focus 'should create a ConfigMap for neutron.conf with' --fail-fast  --procs 1"

Taken from https://github.com/openstack-k8s-operators/nova-operator/pull/303

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,7 @@ PROC_CMD = --procs ${PROCS}
 
 .PHONY: test
 test: manifests generate fmt vet envtest ginkgo ## Run tests.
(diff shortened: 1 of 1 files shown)


mrKisaoLamb (commented):
lgtm

karelyatin:
Noticed it's being done in https://github.com/openstack-k8s-operators/neutron-operator/pull/144, so will close this.

PR #157, opened 2023-06-05
Remove endpoint url and keystone service id from status

kajinamit:
These items presented in CR status are not really used. Endpoints can be discovered using Keystone and we have no use case to require service id.


karelyatin (approved):
/lgtm

PR #167, opened 2023-06-23
Remove HostPath volume mounts

kajinamit:
Timezone can be configured by TZ environment and this is more flexible. (For example users can configure different timezone for pods and platform.)

Also, mounting machine-id does not guarantee unique id in pods because we do not strictly prohibit running multiple pods in the same node.

### DIFF
--- a/pkg/neutronapi/volumes.go
+++ b/pkg/neutronapi/volumes.go
@@ -43,22 +43,6 @@ func GetAPIVolumes(name string, extraVol []neutronv1beta1.NeutronExtraVolMounts,
 
 	res := []corev1.Volume{
 		{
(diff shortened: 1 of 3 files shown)


bshephar (approved):
Sounds good to me.

kajinamit:
/retest

error: Post "https://api.oko-13.oooci.ccitredhat.com:6443/apis/build.openshift.io/v1/namespaces/openstack/buildconfigs": EOF

karelyatin (approved):
/lgtm

Ok so if there is some need to customize TZ, it can be done externally via other options https://access.redhat.com/solutions/2567961

PR #168, opened 2023-06-23
Add support for configuring route annotations

lewisdenny:
This PR adds support for configuring route annotations into the neutron operator and depends on https://github.com/openstack-k8s-operators/lib-common/pull/288

This patch can be tested by editing the NeutronAPI CR with `oc edit NeutronAPI` and adding the following to the spec:
```
  route:
    routeAnnotations:
      haproxy.router.openshift.io/timeout: 10m
```


api/bases/neutron.openstack.org_neutronapis.yaml
```diff
@@ -2092,6 +2092,14 @@ spec:
                       to an implementation-defined value. More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/'
                     type: object
                 type: object
+              routeAnnotations:
+                description: RouteAnnotations to be added to the created route
+                properties:
+                  timeout:
+                    default: "60"
```
lewisdenny, line 2099: Maybe we don't want to have a default at all? 
As a route will already have a 30s timeout from HAProxy, I would like no default and if a timeout isn't provided then no annotation is added

api/v1beta1/neutronapi_types.go
```diff
@@ -194,6 +198,13 @@ type NeutronAPIDebug struct {
 	Service bool `json:"service"`
 }
 
+type NeutronRouteAnnotations struct {
```
lewisdenny, line 201: I can see us having more route annotations in the future so I made this a struct

controllers/neutronapi_controller.go
```diff
@@ -412,12 +412,18 @@ func (r *NeutronAPIReconciler) reconcileInit(
 		ports[metallbcfg.Endpoint] = portCfg
 	}
 
+	routeAnnotations := make(map[string]string)
+	if instance.Spec.RouteAnnotations.Timeout != "" {
```
lewisdenny, line 416: I would have liked to do this with a range but couldn't find an example in the other operators I checked.

controllers/neutronapi_controller.go
```diff
@@ -412,12 +412,18 @@ func (r *NeutronAPIReconciler) reconcileInit(
...
+		routeAnnotations["haproxy.router.openshift.io/timeout"] = instance.Spec.RouteAnnotations.Timeout
```
lewisdenny, line 417: Should `haproxy.router.openshift.io/timeout` be a const?

bshephar: I would avoid using string literals generally, yeah. If we envision that we could potentially need more annotations in the future, we could have a list of consts that contain each like:

```
const (
   routeTimeoutAnnotationName string = "haproxy.router.openshift.io/timeout"
)

Then use the const here instead of the string literal.
```

go.mod
```diff
@@ -88,3 +88,5 @@ require (
 )
 
 replace github.com/openstack-k8s-operators/neutron-operator/api => ./api
+
+replace github.com/openstack-k8s-operators/lib-common/modules/common => /Users/ldenny/Code/openstack/openstack-operators/lib-common/modules/common
```
bshephar, line 92: Reminder that this needs to be removed.

api/v1beta1/neutronapi_types.go
```diff
@@ -194,6 +198,13 @@ type NeutronAPIDebug struct {
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="60"
+	// HAProxy Timeout for route
```
bshephar, line 204: These comments normally start with the name of the variable. Not sure if the linter will complain, but it would normally be like:

```
// Timeout - Provides the HAProxy Timeout value for the route
```

elvgarrui (changes requested):
I think the run lint commit should be squashed with the previous commit in order to have a nice git history.

api/v1beta1/neutronapi_types.go
```diff
@@ -194,6 +198,13 @@ type NeutronAPIDebug struct {
 	Service bool `json:"service"`
 }
 
+// NeutronRouteAnnotations defines the route annotations
+type NeutronRouteAnnotations struct {
+	// +kubebuilder:validation:Optional
+	// Timeout - Provides the HAProxy Timeout value for the route
+	Timeout string `json:"timeout"`
```
booxter, line 205: should we perhaps also choose a sane default value here? (120s?)

api/v1beta1/neutronapi_types.go
```diff
@@ -134,6 +134,10 @@ type NeutronAPISpec struct {
 	// ExtraMounts containing conf files
 	// +kubebuilder:validation:Optional
 	ExtraMounts []NeutronExtraVolMounts `json:"extraMounts,omitempty"`
+
+	// RouteAnnotations to be added to the created route
+	// +kubebuilder:validation:Optional
+	RouteAnnotations NeutronRouteAnnotations `json:"routeAnnotations,omitempty"`
```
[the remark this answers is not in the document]
booxter: @stuggi do you have a link to the topic discussed? thanks.

booxter (changes requested):

booxter:
Please also consider adding a functional test for this knob. Thanks.

api/v1beta1/neutronapi_types.go
bshephar: I’m not sure we should have a default in Neutron-operator. It would be more in line with user expectations that if they don’t provide a value here then the default will be OpenShift route defaults.

If we add a default here, we might cause some confusion for users.

controllers/neutronapi_controller.go
bshephar: I think Lewis is referring to enumerating the struct. But maybe the better option is as you already said, to just use map[string]string so that we aren’t explicit about the annotation provided.

bshephar:
> Please also consider adding a functional test for this knob. Thanks.

Do we start the envtest environment with the Openshift API server? Or do we have some other way to simulate the Route API?

bshephar:
> > Please also consider adding a functional test for this knob. Thanks.
> 
> Do we start the envtest environment with the Openshift API server? Or do we have some other way to simulate the Route API?

Ahh,
https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/test/openshift_crds/README.md

api/v1beta1/neutronapi_types.go
lewisdenny: I think there could be argument here to increase the default... by default. This PR is for editing the server timeout [1] and just taking a quick look at one of the OpenStack labs we have the same value set by default to 10m for Neutron [2]

[1] https://github.com/openshift/router/blob/0cd9a58dff77127562f7851ccfa5cda79b8cef48/images/router/haproxy/conf/haproxy-config.template#L516
[2] timeout server 10m

booxter: We discussed the matter of default choice networking team call just now, and the assumption is that we will retain whatever the default value we have for the same in tripleo, which is 120s at the moment. There was a legitimate reason to bump the value in tripleo for neutron api that we may dig for, but I don't have it at the moment.

That said, I don't think this is required that this PR addresses the choice of the default value. This can be done in a separate PR, if needed.

api/v1beta1/neutronapi_types.go
```diff
@@ -194,6 +198,14 @@ type NeutronAPIDebug struct {
 	Service bool `json:"service"`
 }
 
+// NeutronRoute is used to define all the information for the OpenShift route
```
karelyatin, line 201: This kind of looks duplicating with Names(i.e Route within Route), Route.RouteAnnotations, if it's done to have more similar annotations in future for different resources, couldn't it be better to have like Annotations.Route, Annotations.Service etc? Or Just have all in top level CRD like RouteAnnotations map[string]string, ServiceAnnotations map[string]string?
Not sure if it's already pointed out and i just missed that discussion.

Also please add functional test for this.

karelyatin: Wrt  "NeutronRoute is used to define all the information for the OpenShift route", what all information you mean here? it means just annotations or some other option too?

config/samples/neutron_v1beta1_neutronapi.yaml
```diff
@@ -17,4 +17,4 @@ spec:
   preserveJobs: false
   containerImage: quay.io/podified-antelope-centos9/openstack-neutron-server:current-podified
   replicas: 1
-  secret: neutron-secret
+  secret: neutron-secret
```
karelyatin, line 20: No change in this file so can checkout from previous commit to avoid showing it here.

karelyatin (commented):
Few comments inline

lewisdenny:
Closing in favour of https://github.com/openstack-k8s-operators/lib-common/pull/293

PR #174, opened 2023-07-12
Use consistent commands to copy manifests

kajinamit:
https://github.com/openshift/release/pull/41169 was merged which resolves the permission problem in CI. Thus we can revert the change and use the consistent command.


abays (approved):
/lgtm

PR #196, opened 2023-08-22
[local webhook] Adjust firewall only if binary exists

karelyatin:
The script currently fails on systems where
firewall-cmd binary do not exist. Let's set the
firewall rules only when binary exists.


abays (approved):
/lgtm

PR #201, opened 2023-08-29
Unpin from v0.1.x tags for org dependencies

booxter:
This move requires an adjustment to new api from ovn-operator for a function that was recently replaced.


booxter:
recheck

karelyatin (commented):
/lgtm

PR #203, opened 2023-08-31
Create config entry for metadata agent

booxter:


controllers/neutronapi_controller.go
```diff
@@ -854,6 +858,16 @@ func (r *NeutronAPIReconciler) generateServiceConfigMaps(
...
 
+	// TODO(ihar): make the metadata config secret too
+	templateParameters["TransportURL"] = string(transportURLSecret.Data["transport_url"])
+
+	// TODO(ihar): when is it not localhost?
+	templateParameters["MetadataBindHost"] = "127.0.0.1"
+
+	// TODO(ihar): set to correct values from nova - to figure out how to extract them from nova?
+	templateParameters["NovaMetadataHost"] = "1.2.3.4"
+	templateParameters["MetadataProxySharedSecret"] = "1234567842"
```
gibizer, line 869: While this data coming from the osp-secret both nova- and neutron-operator gets, the password selector is configuration of the Nova CR so this data also depends on that password selector value. As this can be per cell and neutron has no cell concept this secret value should be provided by the same mechanism as the metadata host.

controllers/neutronapi_controller.go
```diff
@@ -854,6 +858,16 @@ func (r *NeutronAPIReconciler) generateServiceConfigMaps(
...
 	templateParameters["SBConnection"] = dbmap["internal-SB"]
 
+	// TODO(ihar): make the metadata config secret too
+	templateParameters["TransportURL"] = string(transportURLSecret.Data["transport_url"])
+
+	// TODO(ihar): when is it not localhost?
+	templateParameters["MetadataBindHost"] = "127.0.0.1"
+
+	// TODO(ihar): set to correct values from nova - to figure out how to extract them from nova?
+	templateParameters["NovaMetadataHost"] = "1.2.3.4"
```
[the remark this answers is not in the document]
mrKisaoLamb: If I recall correctly, in TripleO, we manage the Nova metadata host through the endpoint_data.yaml using NovaMetadataCellInternal. Given the significant changes in our deployment flow with the current design, I agree that providing a secret seems like the easiest approach. One potential improvement I can think of is encapsulating the logic for reading a "universal secret" in lib-common to establish a single location where this kind of secret reader can be defined.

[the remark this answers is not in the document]
gibizer: I pushed a WIP patch on nova-operator to show what I mean https://github.com/openstack-k8s-operators/nova-operator/pull/515

booxter:
The final config as set by neutron-operator:

```
[ihrachys@ansible-fx2-6 edpm-ansible]$ kubectl get secret neutron-external-config -o json | jq '.data | map_values(@base64d)'                                                          
{
  "neutron_ovn_metadata_agent.ini": "[DEFAULT]\ntransport_url = rabbit://default_user_VaHHp0JNfEW7_03TiSg:06H5pilZd6R-SmpSQqrOs4cuWrlwCiFv@rabbitmq.openstack.svc:5672\n\n[ovn]\novn_sb_connection = tcp:172.17.0.31:6642\n"
}
```

This PR still misses functional tests. I will add them when I have the other pieces (dataplane and edpm-ansible) aligned.

templates/neutron_ovn_metadata_agent.ini
```diff
@@ -0,0 +1,11 @@
+# NOTE: some necessary metadata agent options are set by nova.
+# Specifically, nova_metadata_host and metadata_proxy_shared_secret.
+#
+# The final metadata agent deployment will have to merge settings from multiple
+# sources.
+
+[DEFAULT]
+transport_url = {{ .TransportURL }}
```
slawqo, line 8: Do we really need this in neutron-ovn-metadata-agent? I don't think this agent is connecting to the rabbitmq at all.

controllers/neutronapi_controller.go
```diff
@@ -787,6 +787,29 @@ func (r *NeutronAPIReconciler) reconcileNormal(ctx context.Context, instance *ne
...
+		r.Log.Info("No external endpoint defined for SB OVNDBCluster, deleting external Secrets")
+		cleanupSecretsErr := r.deleteExternalSecrets(ctx, helper, instance)
+		if cleanupSecretsErr != nil {
+			r.Log.Error(cleanupSecretsErr, "Failed to delete external Secrets")
+			return ctrl.Result{}, cleanupSecretsErr
+		}
+		return ctrl.Result{}, nil
+	} else {
+		// Create Secrets for external dataplane consumption
+		// TODO(ihar) - is there any hashing mechanism for EDP config? do we trigger deploy somehow?
```
gibizer, line 805: This is a good point. I suggested to @slagle before to:
1) keep the hash of the CM/Secret used for a given deployment of a OpenStackDataPlaneService recorded somewhere (potentially in the Status of the OpenStackDataPlaneService). This way the human operator can compare the hash in the neutron (nova) CR with the hash in the related OpenstackDataPlaneService CR to see if there is a need to redeployment as the config changed.
2) watch the CMs/Secrets the OpenstackDataPlaneService uses and either automatically or with a human acknowledgement do the redeployment of the service if the configuration changes. 

See the future enhancement section of the downstream doc https://docs.google.com/document/d/1tr5KzgT0cB-nCF36kSmEbzxy9i38oLQyfCsGwgQUiY8/edit#heading=h.fxf6yuwcf55m

controllers/neutronapi_controller.go
```diff
@@ -809,6 +832,71 @@ func (r *NeutronAPIReconciler) transportURLCreateOrUpdate(instance *neutronv1bet
...
+		{
+			Name:               fmt.Sprintf("%s-external-config", instance.Name),
+			Namespace:          instance.Namespace,
+			Type:               util.TemplateTypeNone,
+			InstanceType:       instance.Kind,
+			Labels:             cmLabels,
+			ConfigOptions:      templateParameters,
+			AdditionalTemplate: templates,
+		},
+	}
```
gibizer, line 896: Nova also records the hash of the Secret in its Status to serve as a reference point for comparison with what was used for the deployment. See https://github.com/openstack-k8s-operators/nova-operator/pull/515/files#diff-86a1ea2f9281135a424412db58e53334e5e0e4dbf4e2f6b0a315daae0f470999R591

templates/neutron_ovn_metadata_agent.ini
booxter: I didn't realize it doesn't use RPC for any maintenance tasks like agent healthcheck reporting. I will remove this.

controllers/neutronapi_controller.go
```diff
@@ -809,6 +832,71 @@ func (r *NeutronAPIReconciler) transportURLCreateOrUpdate(instance *neutronv1bet
...
+// TODO(ihar) this function could live in lib-common
+// deleteExternalSecrets - delete obsolete secrets for external dataplane consumption
+func (r *NeutronAPIReconciler) deleteExternalSecrets(
+	ctx context.Context,
+	h *helper.Helper,
+	instance *neutronv1beta1.NeutronAPI,
+) error {
+	cm := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-external-config", instance.Name),
```
booxter, line 844: decision to make: whether to have a single "external-config" for all EDP services, or one per service.

booxter: I will have one secret per agent

controllers/neutronapi_controller.go
booxter: Ack, I will leave the todo entry for now.

booxter:
Updated the patch as follows:
- fixed functional test failures due to assumption that hash keys will be updated synchronously with Secret created;
- renamed key in the Secret to 10-neutron-metadata.conf to map the naming used in nova Secret (05-nova-metadata.conf);
- renamed the Secret name to ovn-metadata-agent-neutron-config to map the naming used in nova Secret (nova-metadata-neutron-config).

This should be ready to go in.

slawqo (approved):

karelyatin (approved):
/lgtm

controllers/neutronapi_controller.go
```diff
@@ -809,6 +815,126 @@ func (r *NeutronAPIReconciler) transportURLCreateOrUpdate(instance *neutronv1bet
...
+
+func (r *NeutronAPIReconciler) reconcileExternalMetadataAgentSecret(
+	ctx context.Context,
+	h *helper.Helper,
+	instance *neutronv1beta1.NeutronAPI,
+	envVars *map[string]env.Setter,
+) error {
+	sbCluster, err := ovnclient.GetDBClusterByType(ctx, h, instance.Namespace, map[string]string{}, ovnclient.SBDBType)
+	if err != nil {
+		err = r.deleteExternalSecret(ctx, h, instance, getMetadataAgentSecretName(instance))
```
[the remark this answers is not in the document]
booxter: Not sure about that. But I think what we should do is to eventually introduce a separate status type to track externalConfigs, so that users can check the status in CRD.

PR #204, opened 2023-09-07
Update neutronapi to use service override

stuggi:
Removes creation of routes.Those get done in the openstack-operator. Via service overrides the service can be customized. The service operator adds annotation to the service which needs to be exposed as a route.

Jira: [OSP-26690](https://issues.redhat.com//browse/OSP-26690)

Depends-On: openstack-k8s-operators/lib-common#332


abays (approved):
/lgtm

PR #210, opened 2023-09-18
Create config Secret for DHCP agent

booxter:
The config is (right now) exactly the same as for sriov agent. But we still want to keep the secrets separated, if nothing else - to keep key names and secret names meaningful and explicit.


booxter:
@slawqo I know that we don't have ansible piece for this yet, but I think it would make sense to merge this as a start. I am not sure if anything but transport_url will be needed anyway. But in case there are additional pieces to include, we can always expand the config template.

slawqo (approved):

PR #213, opened 2023-09-25
[envtest]Use keystone helpers from keystone-operator

gibizer:
This was necessary to remove a dependency cycle from lib-common


stuggi (approved):
/lgtm

PR #215, opened 2023-09-28
[envtest]Use database helpers from mariadb-operator

gibizer:
This is necessary to remove a dependency cycle from lib-common


karelyatin (approved):
/lgtm

PR #217, opened 2023-09-29
Enable caching using memcached in the neutron server

slawqo:
Closes: [OSP-26502](https://issues.redhat.com//browse/OSP-26502)

### DIFF
--- a/controllers/neutronapi_controller.go
+++ b/controllers/neutronapi_controller.go
@@ -19,17 +19,21 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	memcachedv1 "github.com/openstack-k8s-operators/infra-operator/apis/memcached/v1beta1"
 	rabbitmqv1 "github.com/openstack-k8s-operators/infra-operator/apis/rabbitmq/v1beta1"
 	keystonev1 "github.com/openstack-k8s-operators/keystone-operator/api/v1beta1"
 	"github.com/openstack-k8s-operators/lib-common/modules/common"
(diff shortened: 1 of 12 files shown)


controllers/neutronapi_controller.go
```diff
@@ -720,6 +727,36 @@ func (r *NeutronAPIReconciler) reconcileNormal(ctx context.Context, instance *ne
...
+		return ctrl.Result{}, err
+	}
+
+	if !memcached.IsReady() {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.MemcachedReadyCondition,
+			condition.RequestedReason,
+			condition.SeverityInfo,
+			condition.MemcachedReadyWaitingMessage))
+		return ctrl.Result{RequeueAfter: time.Duration(10) * time.Second}, fmt.Errorf("memcached %s is not ready", memcached.Name)
```
fmount, line 755: Overall the change itself is a good one and I don't have anything against it. 
I understand that we're going to introduce the dependency to a memcached instance, which means that the deployment won't progress if it's not present in the main ctlplane.
If we're happy with this config, then +1 from me, otherwise we need to improve the logic a bit and make it optional (e.g., if I don't specify a memcached instance to watch for, don't block the neutron operator deployment).

fmount: and looking at TripleO this change seems consistent as we used to enable memcached by default as part of the regular overcloud roles [1]

[1] https://github.com/openstack/tripleo-heat-templates/blob/master/overcloud-resource-registry-puppet.j2.yaml#L163

slawqo: This is also consistent with at least heat and keystone operators: https://github.com/openstack-k8s-operators/keystone-operator/blob/main/controllers/keystoneapi_controller.go#L702 and https://github.com/openstack-k8s-operators/heat-operator/blob/main/controllers/heat_controller.go#L387 so I think we can/should keep it like it is proposed now. Any objections?

fmount: Not from me, I'm +1

fmount (commented):
/lgtm

fmount:
Other than the conflict I see w/ `go.mod`, I suspect we should first solve the kuttl issue [1]:

```
case.go:366: resource Pod:neutron-kuttl-tests/: .metadata.annotations.openshift.io/scc: value mismatch, expected: anyuid != actual: restricted-v2
```

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_neutron-operator/217/pull-ci-openstack-k8s-operators-neutron-operator-main-neutron-operator-build-deploy-kuttl/1712420408214425600/artifacts/neutron-operator-build-deploy-kuttl/openstack-k8s-operators-kuttl/build-log.txt

controllers/neutronapi_controller.go
karelyatin: Yes it would have been good to have this optional but as other base services depends on memcache already we can't avoid it fully. So optional part can be done later with other services if felt desired.

karelyatin (commented):
/lgtm

PR #221, opened 2023-10-04
Remove kolla bits from service configuration path

booxter:


slawqo (approved):

SeanMooney:
this is in direct contoduction to the agreed approach we had selected in 
https://github.com/openstack-k8s-operators/docs/blob/main/service_config.md for all services
services running on openshfit.

PR #223, opened 2023-10-05
Run neutronapi / db-sync as neutron:neutron

booxter:
Now that we removed kolla bits from the startup path, we can unprivilege the containers.

Note: since we still may mount HostPaths, we have to stick to hostmount-anyuid for now. If this interface is revisited and e.g. switched to PVCs, then we should be able to tighten SCC a bit more.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -204,7 +204,7 @@ rules:
   - security.openshift.io
   resourceNames:
   - anyuid
-  - privileged
+  - hostmount-anyuid
   resources:
   - securitycontextconstraints
   verbs:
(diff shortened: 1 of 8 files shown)


booxter:
/hold

looks like we will have to reconcile this (and kolla removal) with tls folks plans

booxter:
@stuggi could you please clarify if TLS squad found a way to build CA bundles without root / kolla? I'd like to move forward with this neutron (and upcoming ovn) PRs to unprivilege networking services, but I'm holding the work for now to understand if this is the path to proceed, or we'll have to revert to kolla/root for some reason. Thanks!

stuggi:
@booxter yes we have a solution. we create the full CA bundle for the deployments and can mount this to the env ca trust location. so no need to copy or update the CA trust via kolla

karelyatin (commented):
/lgtm

PR #228, opened 2023-10-13
Update tenant_network_types to allow other types to be created

karelyatin:
Currently only geneve tenant networks are allowed, open up it to allow other network types to be created by default.

### DIFF
--- a/templates/neutronapi/config/01-neutron.conf
+++ b/templates/neutronapi/config/01-neutron.conf
@@ -16,13 +16,19 @@ connection=mysql+pymysql://{{ .DbUser }}:{{ .DbPassword }}@{{ .DbHost }}/{{ .Db 
 [ml2]
 mechanism_drivers = ovn
 type_drivers = local,flat,vlan,geneve,vxlan
-tenant_network_types = geneve
+tenant_network_types = geneve,vxlan,vlan,flat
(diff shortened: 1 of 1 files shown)


booxter:
We should definitely __be able to__ allow users access to other network types. I wonder if this discussion should be split into:

- what's the best default? (probably a good idea to check in tripleo - though I don't know if they have a single recommendation - a link shared to show it would be helpful)
- if this should be configurable (I think so - it was in tripleo and I think it's needed to allow admins to control which types tenants can use) - do you think it's a separate topic from this PR?

booxter:
As per Yatin and official RH docs,

```
Open Virtual Networking (OVN) is the default networking mechanism driver in Red Hat OpenStack Platform {vernum}. If you want to use OVN with distributed virtual routing (DVR), you must include the `environments/services/neutron-ovn-dvr-ha.yaml` file in the `openstack overcloud deploy` command. If you want to use OVN without DVR, you must include the `environments/services/neutron-ovn-ha.yaml` file in the `openstack overcloud deploy` command.
```

Both files list all four as default types. So it's a sane default.

Override part should be possible with a customConfig snippet, so probably also not an issue.

gouthampacha:
+1 

Thanks @karelyatin

PR #235, opened 2023-10-23
Enable mysql_wsrep_sync_wait galera db option

karelyatin:
It is required to be set for multi master galera, without it set there can be reads from not up to date db instance and that leads to various issues.


abays (approved):
/lgtm

karelyatin:

ovn images refreshed

PR #246, opened 2023-11-10
Add httpd sidecontainer for the Neutron API

slawqo:
This patch introduces an additional sidecontainer for the NeutronAPI that is
supposed to ProxyPass requests to the neutron-api process behind. This will also solve the problem of tls-everywhere for Neutron, where the second segment is re-encrypted from the Route (created by the openstack operator) to the deployment.

* The httpd will serve requests on 9696, while the neutron-api service will be run locally on 9697
* httpd instance will be customized in a follow up patch to mount certificates and add the related ssl directives

Jira: [OSP-29184](https://issues.redhat.com//browse/OSP-29184)

### DIFF
--- a/pkg/neutronapi/deployment.go
+++ b/pkg/neutronapi/deployment.go
@@ -25,7 +25,8 @@ import (
 )
 
 const (
-	ServiceCommand = "/usr/bin/neutron-server"
+	ServiceCommand         = "/usr/bin/neutron-server"
+	NeutronAPIHttpdCommand = "/usr/sbin/httpd"
 )
 
 // Deployment func
@@ -49,6 +50,7 @@ func Deployment(
 	}
 	cmd := ServiceCommand
 	args := []string{}
+	httpd_args := []string{"-DFOREGROUND"}
 
 	if instance.Spec.Debug.Service {
 		cmd = "/bin/sleep"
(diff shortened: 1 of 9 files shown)


pkg/neutronapi/scc.go
```diff
@@ -20,3 +20,11 @@ func getNeutronSecurityContext() *corev1.SecurityContext {
 		},
 	}
 }
+
+func getNeutronHttpdSecurityContext() *corev1.SecurityContext {
+	runAsUser := int64(0)
```
booxter, line 25: can't it be "apache" user id here?

booxter:
This seems fine by me. I have a question about user ID used, but we may follow up on it later if needed.

pkg/neutronapi/deployment.go
```diff
@@ -112,6 +114,19 @@ func Deployment(
 							LivenessProbe:            livenessProbe,
```
karelyatin, line 114: shouldn't the probes be changed now to the new ports?

templates/neutronapi/httpd/httpd.conf
```diff
@@ -0,0 +1,23 @@
+ServerTokens Prod
+ServerSignature Off
+TraceEnable Off
+ServerRoot "/etc/httpd"
+ServerName "neutron.openstack.svc"
```
karelyatin, line 5: seems openstack here is namespace name, and neutron CR name, we shouldn't hard code these

karelyatin (commented):
sorry got late on this, overall looks good, just few confirmation inline

PR #257, opened 2023-11-28
    Reintroduce kolla into startup path for neutron-api

booxter:
This is to keep neutron-api in line with [1] until we decide otherwise.

Keeping kolla has several drawbacks, specifically:

1. allowPrivilegeEscalation has to be true for sudo.
2. dropCapabilities cannot be ALL because kolla_set_configs failed to change to root gid.

This patch may be revisited later if we decide to tighten privileges further by patching our kolla fork, or removing kolla completely from the startup path. For now, we'll have to stick to common approach.

The discussion over kolla place can be continued in jira [2].

[1] https://github.com/openstack-k8s-operators/docs/blob/main/service_config.md
[2] https://issues.redhat.com/browse/OSPRH-2240

### DIFF
--- a/pkg/neutronapi/dbsync.go
+++ b/pkg/neutronapi/dbsync.go
@@ -1,6 +1,7 @@
 package neutronapi
 
 import (
+	"github.com/openstack-k8s-operators/lib-common/modules/common/env"
 	neutronv1beta1 "github.com/openstack-k8s-operators/neutron-operator/api/v1beta1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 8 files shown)


api/bases/neutron.openstack.org_neutronapis.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.11.1
```
slawqo, line 6: why this change?

config/crd/bases/neutron.openstack.org_neutronapis.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.11.1
```
slawqo, line 6: and same here

karelyatin:
lgtm, the failures are related to patch so needs to be fixed

karelyatin (commented):
/lgtm

PR #263, opened 2023-12-08
[tlse] tls for NeutronAPI pod configuration

stuggi:
Public/Internal service cert secrets and the CA bundle secret can be passed to configure httpd virtual hosts for tls termination. The certs get direct mounted to the appropriate place in etc/pki/tls/certs/%s.crt|key and a CA bundle to
/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem . Job deployments for bootstrap/cron get the CA bundle added if configured.

Also indexes the named input resources for password, CA bundle, and endpoint secrets to be able to watch them for any changes and reconcile if needed.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/428

Jira: [OSPRH-2197](https://issues.redhat.com//browse/OSPRH-2197)


stuggi:
recheck

templates/neutronapi/httpd/10-neutron-httpd.conf
```diff
@@ -1,4 +1,8 @@
+{{ range $endpt, $vhost := .VHosts }}
+# {{ $endpt }} vhost {{ $vhost.ServerName }} configuration
 <VirtualHost *:9696>
```
slawqo, line 3: shouldn't we use port 13696 if TLS is enabled? IIRC this is what we are using in e.g. OSP 17.1

stuggi: why you think we should do this? its just an internal port on the httpd behind a k8s service.

slawqo: ok, so in such case we don't need to do it like that. You are right

slawqo (approved):

stuggi:
@slawqo I have added some more tests and rebased it. can you please restore the approval?

karelyatin (approved):

PR #288, opened 2024-01-26
[go.mod]Bump from v0.3.0 to pseudoversion

gibizer:
Our intention is to track service operator and lib-common dependencies via pseudoversions. However renovate does not automatically bump from a tagged version (e.g. v0.3.0) to a newer but not tagged pseudoversion. So this patch does the manual bump. After this renovate will bump the newer pseduoversions.


booxter:
The failure is probably related to https://github.com/openstack-k8s-operators/mariadb-operator/pull/172/files that added support for charsets for databases. But it's supposed to default to utf8, but the output suggests it's "".

slawqo (approved):

PR #313, opened 2024-02-28
Add kuttl test for neutron operator: change neutron debug config

brfrenkel:
Implement a kuttl test to ensure that any modifications to the neutron service settings lead to the creation of a new pod with the updated settings. These tests should first confirm that the default logging setting excludes debug details. Then, after adjusting the neutron configuration to enable debug mode (debug = true), the tests should verify that the logs from the newly created pod now contain debug information.

### DIFF
--- a/test/kuttl/common/scripts/check_debug_in_neutron_pod_logs.sh
+++ b/test/kuttl/common/scripts/check_debug_in_neutron_pod_logs.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+pod=$(oc get pods -n $NAMESPACE -l service=neutron -o name)
+
+# Check if the neutron pod logs contain DEBUG messages
+oc logs -n $NAMESPACE "$pod" | grep -q "DEBUG"
+exit $?
--- a/test/kuttl/tests/change_neutron_config/02-assert.yaml
+++ b/test/kuttl/tests/change_neutron_config/02-assert.yaml
@@ -1,0 +1,6 @@
+# check that by default, debug is set in neutron config
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+  - script: |
+        $NEUTRON_KUTTL_DIR/../common/scripts/check_debug_in_neutron_pod_logs.sh
--- a/test/kuttl/tests/change_neutron_config/03-assert.yaml
+++ b/test/kuttl/tests/change_neutron_config/03-assert.yaml
@@ -1,0 +1,7 @@
+# check that now, debug is not set in neutron config
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+  - script: |
+      $NEUTRON_KUTTL_DIR/../common/scripts/check_debug_in_neutron_pod_logs.sh
+      test $? -ne 0
--- a/test/kuttl/tests/change_neutron_config/03-change_neutron_debug_config.yaml
+++ b/test/kuttl/tests/change_neutron_config/03-change_neutron_debug_config.yaml
@@ -1,0 +1,8 @@
+apiVersion: neutron.openstack.org/v1beta1
+kind: NeutronAPI
+metadata:
+  name: neutron
+spec:
+  customServiceConfig: |
+    [DEFAULT]
+    debug = false
--- a/test/kuttl/tests/change_neutron_config/01-assert.yaml
+++ b/test/kuttl/tests/change_neutron_config/01-assert.yaml
@@ -1,0 +1,1 @@
+../../common/assert_sample_deployment.yaml
--- a/test/kuttl/tests/change_neutron_config/01-deploy-neutron.yaml
+++ b/test/kuttl/tests/change_neutron_config/01-deploy-neutron.yaml
@@ -1,0 +1,6 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
+      cp ../../../../config/samples/neutron_v1beta1_neutronapi.yaml deploy
+      oc kustomize deploy | oc apply -n $NAMESPACE -f -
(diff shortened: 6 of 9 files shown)


test/kuttl/tests/change_neutron_config/03-change_neutron_debug_config.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: neutron.openstack.org/v1beta1
+kind: NeutronAPI
+metadata:
+  name: neutron
+  namespace: openstack
```
cescgina, line 5: the namespace line should be removed, since we run the kuttl tests in a different namespace

brfrenkel: fixed, thank you.

brfrenkel:
The kuttl job passed. I also test it manually in a k8s env.
Appreciate your review:)

test/kuttl/common/scripts/check_debug_in_neutron_pod_logs.sh
```diff
@@ -0,0 +1,20 @@
...
+if [ "$1" = "--reverse" ];then
+    # check there is no DEBUG line in the logs
+    found=1
+    not_found=0
+fi
+
+pod=$(oc get pods -n $NAMESPACE -l service=neutron -o name)
+# in the case the logs have DEBUG lines, keep only one to avoid cluttering the
+# test output
+debug=$(oc logs -n $NAMESPACE "$pod" | grep "DEBUG" | head -n 1)
```
slawqo, line 14: You can do something like:
```
oc logs -n $NAMESPACE "$pod" | grep -q "DEBUG"
if [ $? -eq 0 ]; then
    exit $found
else
    exit $not_found
fi
```

test/kuttl/common/scripts/check_debug_in_neutron_pod_logs.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+#
+found=0
+not_found=1
+if [ "$1" = "--reverse" ];then
+    # check there is no DEBUG line in the logs
+    found=1
+    not_found=0
+fi
```
slawqo, line 9: instead of changing value for $found and $not_found You can simple pass "-v" to the grep command if You are expecting to not found the DEBUG line.

test/kuttl/tests/change_neutron_config/02-assert.yaml
```diff
@@ -0,0 +1,5 @@
+# check that by default, debug is set in neutron config
```
slawqo, line 1: why this file has name "02-assert" and not "01-assert"?

brfrenkel: we have already 01-assert file, that uses the assert_sample_deployment. 02-assert checking only the debug config (without 02-action_file)

test/kuttl/common/scripts/check_debug_in_neutron_pod_logs.sh
brfrenkel: let me change and test it please.

test/kuttl/common/scripts/check_debug_in_neutron_pod_logs.sh
brfrenkel: Do you mean that we can use the --reverse option to detect if we want it to be found or not?

test/kuttl/common/scripts/check_debug_in_neutron_pod_logs.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+#
+found=0
+not_found=1
+if [ "$1" = "--reverse" ];then
```
booxter, line 5: nit: add space before `then`

test/kuttl/common/scripts/check_debug_in_neutron_pod_logs.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+#
+found=0
```
booxter, line 3: you definitely don't need two flags for the same state. pick one and calculate the other as needed.

test/kuttl/tests/change_neutron_config/03-assert.yaml
```diff
@@ -0,0 +1,5 @@
+# check that now, debug is not set in neutron config
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+  - script: $NEUTRON_KUTTL_DIR/../common/scripts/check_debug_in_neutron_pod_logs.sh --reverse
```
booxter, line 5: I'd suggest to not make determination of what to check here; instead, leave the decision to the caller. you should be able to put `script` command as follows:

for test with debug=false:

`test $(check_debug_in_neutron_pod_logs.sh) = 0`

for test with debug=true:

`test $(check_debug_in_neutron_pod_logs.sh) != 0`

Then you can remove --reverse flag; and you can just return the result of `grep` from your script.

test/kuttl/common/scripts/check_debug_in_neutron_pod_logs.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+#
```
booxter, line 2: nit: remove # from the line, it's of no use

test/kuttl/common/scripts/check_debug_in_neutron_pod_logs.sh
slawqo: yes, I was thinking about this option exactly :)

brfrenkel:
Thank you folks for your great points! I made the changes and it's really looks better now:)
I cant find the reason the neutron-operator-build-deploy-kuttl is failing, but I saw this failure also in other PRs.
Anyone can point to the reason?

test/kuttl/tests/change_neutron_config/03-assert.yaml
booxter: To reflect discussion elsewhere, you should use `-ne` and `-eq` since these are integers.

slawqo (approved):

PR #314, opened 2024-02-28
Add NeutronAPISpecCore struct

dprince:
This version of the struct (called "core") is meant to be used via the OpenStackControlplane. It is the same as NeutronAPISpec only it is missing the containerImages.

The Default() function for webhooks has been updated accordingly.

Jira: [OSPRH-4835](https://issues.redhat.com//browse/OSPRH-4835)


karelyatin (commented):
/lgtm

PR #316, opened 2024-03-04
Use bind_host=localhost

rabi:
Using localhost would allow bind_host to use the correct loopback address for ipv6.


hjensas (approved):

karelyatin (approved):
/lgtm

PR #320, opened 2024-03-11
Use common memcached.GetMemcachedServerListWithInetString()

stuggi:
Use common func to retrieve the memcached server list for a memcached instance.


olliewalsh (approved):
Look good to me but will wait for Ihar to review

PR #324, opened 2024-03-22
Add default timeout for route annotations

elvgarrui:
Add the timeout property as required annotation for the Neutron API Route. This needs to be followed by a commit in openstack-operator since the route is defined there. In that commit the GetDefaultRouteAnnotations method is used accordingly.

(Goes together with https://github.com/openstack-k8s-operators/openstack-operator/pull/722)

### DIFF
--- a/api/v1beta1/neutronapi_types.go
+++ b/api/v1beta1/neutronapi_types.go
@@ -266,7 +266,8 @@ func (instance NeutronAPI) RbacResourceName() string {
 func SetupDefaults() {
 	// Acquire environmental defaults and initialize Neutron defaults with them
 	neutronDefaults := NeutronAPIDefaults{
-		ContainerImageURL: util.GetEnvVar("RELATED_IMAGE_NEUTRON_API_IMAGE_URL_DEFAULT", NeutronAPIContainerImage),
+		ContainerImageURL:      util.GetEnvVar("RELATED_IMAGE_NEUTRON_API_IMAGE_URL_DEFAULT", NeutronAPIContainerImage),
+		NeutronAPIRouteTimeout: "120s",
 	}
 
 	SetupNeutronAPIDefaults(neutronDefaults)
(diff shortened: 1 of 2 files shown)


api/v1beta1/neutronapi_webhook.go
```diff
@@ -69,6 +70,8 @@ func (spec *NeutronAPISpec) Default() {
 	// only container image validations go here
 	if spec.ContainerImage == "" {
 		spec.ContainerImage = neutronAPIDefaults.ContainerImageURL
+	if spec.NeutronAPIRouteTimeout == "" {
```
dprince, line 73: You need to put this validation in the *core* validations block below for it to apply via the OpenStackControlplane CR.

api/v1beta1/neutronapi_types.go
```diff
@@ -135,6 +135,10 @@ type NeutronAPISpecCore struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// TLS - Parameters related to the TLS
 	TLS NeutronApiTLS `json:"tls,omitempty"`
+
+	// +kubebuilder:validation:Required
+	// NeutronAPI Route Timeout (will be set to environmental default if empty)
+	NeutronAPIRouteTimeout string `json:"neutronAPIRouteTimeout"`
```
booxter, line 141: I wonder if need the field at all. I think with the default values expanded to include the necessary annotations, openstack-operator already sets 120s for the route object via override field. And a user then can go and edit the value for a different one, if needed (I tried with 130s and it worked). On the other hand, this field seems to be ignored by openstack-operator.

Why is setting a custom value through override section for the route in CP CR not enough?

api/v1beta1/neutronapi_webhook.go
elvgarrui: I think this is not longer needed since I removed the spec.NeutronAPIRouteTimeout value. Thanks for the input!

api/v1beta1/neutronapi_types.go
elvgarrui: Done!

booxter:
This looks good now!

karelyatin:
recheck cert-manager issue fixed

karelyatin (approved):
/lgtm

PR #328, opened 2024-04-01
Re-init conditions each reconcile

fmount:
This patch follows the same pattern applied to the other operators, where we re-init the condition at each reconcile loop. `Conditions` are re-evaluated and updated, keeping the `LastTransitionTime` for those that haven't changed (it avoids the transition from True to Unknown to True again).
In addition, the `observedGeneration` field is introduced, and it is used by the `openstack-operator` to check the `IsReady()` function for a particular CR in case a minor update is triggered. All the conditions are evaluated during the main Reconcile loop ( or the `reconcileNormal` function in some circumstances), hence the main `ReadyCondition` is updated within the same flow.
The defer function still updates the `Resource` and mirrors the condition to the top-level `CR`.

Jira: [OSPRH-5919](https://issues.redhat.com//browse/OSPRH-5919)
Jira: [OSPRH-5698](https://issues.redhat.com/browse/OSPRH-5698)

### DIFF
--- a/controllers/neutronapi_controller.go
+++ b/controllers/neutronapi_controller.go
@@ -134,17 +134,23 @@ func (r *NeutronAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		return ctrl.Result{}, err
 	}
 
-	// Always patch the instance status when exiting this function so we can persist any changes.
+	// initialize status if Conditions is nil, but do not reset if it already
+	// exists
+	isNewInstance := instance.Status.Conditions == nil
+	if isNewInstance {
+		instance.Status.Conditions = condition.Conditions{}
+	}
+
+	// Save a copy of the condtions so that we can restore the LastTransitionTime
+	// when a condition's state doesn't change.
+	savedConditions := instance.Status.Conditions.DeepCopy()
+
+	// Always patch the instance status when exiting this function so we can
+	// persist any changes.
 	defer func() {
-		// update the Ready condition based on the sub conditions
-		if instance.Status.Conditions.AllSubConditionIsTrue() {
-			instance.Status.Conditions.MarkTrue(
-				condition.ReadyCondition, condition.ReadyMessage)
-		} else {
-			// something is not ready so reset the Ready condition
-			instance.Status.Conditions.MarkUnknown(
-				condition.ReadyCondition, condition.InitReason, condition.ReadyInitMessage)
-			// and recalculate it based on the state of the rest of the conditions
+		condition.RestoreLastTransitionTimes(
+			&instance.Status.Conditions, savedConditions)
+		if instance.Status.Conditions.IsUnknown(condition.ReadyCondition) {
 			instance.Status.Conditions.Set(
 				instance.Status.Conditions.Mirror(condition.ReadyCondition))
 		}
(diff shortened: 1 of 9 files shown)


controllers/neutronapi_controller.go
```diff
@@ -448,8 +449,21 @@ func (r *NeutronAPIReconciler) reconcileInit(
...
+				err.Error()))
 			return ctrlResult, err
 		} else if (ctrlResult != ctrl.Result{}) {
+			// Marking the condition as Unknown because we are not returining
+			// an err, but comparing the ctrlResult: this represents an in
+			// progress operation rather than something that failed
+			instance.Status.Conditions.MarkUnknown(
+				condition.TLSInputReadyCondition,
+				condition.InitReason,
+				condition.InputReadyInitMessage)
```
abays, line 466: I think it would be better to use these [1] [2] here.

[1] https://github.com/openstack-k8s-operators/lib-common/blob/56015b1ae3f62721195798f7564f0abcba937547/modules/common/condition/conditions.go#L86-L87
[2] https://github.com/openstack-k8s-operators/lib-common/blob/56015b1ae3f62721195798f7564f0abcba937547/modules/common/condition/conditions.go#L137-L138

controllers/neutronapi_controller.go
```diff
@@ -461,8 +475,21 @@ func (r *NeutronAPIReconciler) reconcileInit(
...
+			err.Error()))
 		return ctrlResult, err
 	} else if (ctrlResult != ctrl.Result{}) {
+		// Marking the condition as Unknown because we are not returining
+		// an err, but comparing the ctrlResult: this represents an in
+		// progress operation rather than something that failed
+		instance.Status.Conditions.MarkUnknown(
+			condition.TLSInputReadyCondition,
+			condition.InitReason,
+			condition.InputReadyInitMessage)
```
abays, line 492: Same thought regarding the reason and message here

controllers/neutronapi_controller.go
fmount: Makes sense, updating that!

controllers/neutronapi_controller.go
fmount: ++

controllers/neutronapi_controller.go
```diff
@@ -448,8 +449,21 @@ func (r *NeutronAPIReconciler) reconcileInit(
...
+				condition.TLSInputErrorMessage,
+				err.Error()))
 			return ctrlResult, err
 		} else if (ctrlResult != ctrl.Result{}) {
+			// Marking the condition as Unknown because we are not returining
+			// an err, but comparing the ctrlResult: this represents an in
+			// progress operation rather than something that failed
+			instance.Status.Conditions.MarkUnknown(
+				condition.TLSInputReadyCondition,
+				condition.InitReason,
```
abays, line 465: Could we also use `RequestedReason` here and on L491 instead?  It seems to be what we use for these sorts of situations across the operators.  For example [1].

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/c47a8894eea5c6a5a7354e53f913a61032012835/controllers/keystoneapi_controller.go#L720-L722

fmount: yeah, makes sense, updating in my next PS

slawqo (approved):

PR #329, opened 2024-04-02
Add *core* validation for default route annotations

dprince:
Add a *core* API spec version of the new
GetDefaultRouteAnnotations function. The new
OpenStackControlPlane PR will require use of this
version in order to leverage the same function.


karelyatin (commented):
/lgtm

elvgarrui:
Thanks for catching this! Looks good to me.

PR #335, opened 2024-04-12
Check Deployment ObservedGeneration to set ReadyCondition

fmount:
`DeploymentReadyCondition` is something that depends on the number of ready `Replicas`, but it should also be influenced by the `ObservedGeneration`. We need to make sure that the readiness is set for the last `CR` version, and this patch adds the additional check that we're rolling out to the other operators as well.

### DIFF
--- a/controllers/neutronapi_controller.go
+++ b/controllers/neutronapi_controller.go
@@ -1009,34 +1009,38 @@ func (r *NeutronAPIReconciler) reconcileNormal(ctx context.Context, instance *ne
 		return ctrlResult, nil
 	}
 
-	instance.Status.ReadyCount = depl.GetDeployment().Status.ReadyReplicas
+	if depl.GetDeployment().Generation <= depl.GetDeployment().Status.ObservedGeneration {
+		instance.Status.ReadyCount = depl.GetDeployment().Status.ReadyReplicas
 
-	// verify if network attachment matches expectations
-	networkReady, networkAttachmentStatus, err := nad.VerifyNetworkStatusFromAnnotation(ctx, helper, instance.Spec.NetworkAttachments, serviceLabels, instance.Status.ReadyCount)
-	if err != nil {
-		return ctrl.Result{}, err
-	}
+		// verify if network attachment matches expectations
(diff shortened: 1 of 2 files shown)


controllers/neutronapi_controller.go
```diff
@@ -1032,8 +1032,12 @@ func (r *NeutronAPIReconciler) reconcileNormal(ctx context.Context, instance *ne
 		return ctrl.Result{}, err
 	}
 
-	if instance.Status.ReadyCount > 0 {
-		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
+	if instance.Status.ReadyCount > 0 &&
+		depl.GetDeployment().Generation <= depl.GetDeployment().Status.ObservedGeneration {
```
fmount, line 1036: @mrkisaolamb as per [1] I think this is the right check to do, right?

[1] https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/util/deployment_util.go#L712

mrkisaolamb: @fmount 
Yes, this seems to be the most common approach for guarding against unexpected resets and re-creations

controllers/neutronapi_controller.go
```diff
@@ -1009,43 +1009,46 @@ func (r *NeutronAPIReconciler) reconcileNormal(ctx context.Context, instance *ne
...
+				err.Error()))
 
-		return ctrl.Result{}, err
-	}
+			return ctrl.Result{}, err
+		}
 
-	if instance.Status.ReadyCount > 0 {
-		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
+		if instance.Status.ReadyCount > 0 {
```
fmount, line 1036: we'll change this to make sure Spec.Replicas matches the ReadyCount in a different PR.

fmount:
recheck

dprince (approved):

fmount:
recheck

PR #341, opened 2024-04-22
[tlse] Ensure input readiness reported after all certificates are ready

Deydra71:
Fixed a missing condition that prevented the deployment from being marked as ready, even though all certificates and attached secrets were properly prepared.

### DIFF
--- a/controllers/neutronapi_controller.go
+++ b/controllers/neutronapi_controller.go
@@ -495,6 +495,9 @@ func (r *NeutronAPIReconciler) reconcileInit(
 	}
 	secretVars[tls.TLSHashName] = env.SetValue(certsHash)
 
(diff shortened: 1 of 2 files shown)


stuggi:
can you add at the top of https://github.com/openstack-k8s-operators/neutron-operator/blob/main/test/functional/neutronapi_controller_test.go#L1071 a validation that the `TLSInputCondition` is true?

```
			th.ExpectCondition(
				neutronAPIName,
				ConditionGetterFunc(NeutronAPIConditionGetter),
				condition.TLSInputReadyCondition,
				corev1.ConditionTrue,
			)
```

stuggi (approved):
/lgtm

PR #349, opened 2024-05-17
style: Uncapitalize error messages

booxter:
As per [1], errors should be uncapitalized, "since they are usually printed following other context".

While at it, replaced Ovn -> OVN in several error messages.

1: https://go.dev/wiki/CodeReviewComments


karelyatin (approved):
/lgtm

PR #351, opened 2024-05-17
style: Flip if-else order where it saves indent

booxter:
As per [1], we should avoid unnecessary indents in these cases.

```
Try to keep the normal code path at a minimal indentation, and indent
the error handling, dealing with it first.
```

1: https://go.dev/wiki/CodeReviewComments


karelyatin (commented):
/lgtm

PR #395, opened 2024-08-12
Add project_name to nova/placement client config

karelyatin:
With missing 'project_name' config in adoption jobs nova api service requests fails with [1] as the service users created[2] in adopted cloud don't have default_project_id set. And user is also not updated if it already exists[3].

It's not seen in the non-adoption jobs as users created have default_project_id set[3]. This patch explicitly set's the project_name config to resolve the issue.

[1] keystoneauth1.exceptions.catalog.EmptyCatalog: The service catalog is empty
[2] https://github.com/openstack-archive/tripleo-ansible/blob/stable/wallaby/tripleo_ansible/roles/tripleo_keystone_resources/tasks/users.yml#L40
[3] https://github.com/openstack-k8s-operators/lib-common/blob/719002a/modules/openstack/user.go#L55-L56
[4] https://github.com/openstack-k8s-operators/lib-common/blob/719002a/modules/openstack/user.go#L66

Related-Issue: [OSPCIX-404](https://issues.redhat.com//browse/OSPCIX-404)


karelyatin:

quay.rdoproject.org issue

gibizer (approved):
make sense to me. The nova-operator also sets this.

PR #403, opened 2024-08-29
improvement(neutronapi): Enhance readiness probe to verify SSL certificate presenceAdd 

son-vyas:
improvement(neutronapi): Enhance readiness probe to verify SSL certificate presence

- Updated the existing readiness probe in the Neutron API deployment to check for the presence of the internal.crt SSL certificate before marking the pod as ready.
- Applied the SSL certificate check to the HTTPD container as well, ensuring secure traffic handling.
- Improved the deployment reliability by ensuring that pods are only marked as ready when they have the necessary SSL certificates in place.


slawqo:
This requires more work as CI failures looks like definitely related to that patch

PR #406, opened 2024-09-05
Add kuttl tests to validate config overrides with extraMounts

karelyatin:
Added config override check for both secret and configmap.

Also update debug config check pod to filter only running containers.

Resolves: [OSPRH-4296](https://issues.redhat.com//browse/OSPRH-4296)


karelyatin:

quay.rdo disabled

PR #409, opened 2024-09-10
Untangle service from ovn requirement

booxter:
Note to reviewers: I advise to ignore white space changes when reviewing Functional test changes.

### DIFF
--- a/api/bases/neutron.openstack.org_neutronapis.yaml
+++ b/api/bases/neutron.openstack.org_neutronapis.yaml
@@ -2052,6 +2052,12 @@ spec:
                 default: memcached
                 description: Memcached instance name.
                 type: string
+              ml2Drivers:
+                description: Ml2Drivers - list of ml2 drivers to enable. Using ["ovn"]
+                  if not set.
+                items:
+                  type: string
+                type: array
               networkAttachments:
                 description: NetworkAttachments is a list of NetworkAttachment resource
                   names to expose the services to the given network
--- a/api/bases/neutron.openstack.org_neutronapis.yaml
+++ b/api/bases/neutron.openstack.org_neutronapis.yaml
@@ -2053,8 +2053,10 @@ spec:
                 description: Memcached instance name.
                 type: string
               ml2MechanismDrivers:
+                default:
+                - ovn
                 description: Ml2MechanismDrivers - list of ml2 drivers to enable.
-                  Using ["ovn"] if not set.
+                  Using {"ovn"} if not set.
                 items:
                   type: string
                 type: array
--- a/api/bases/neutron.openstack.org_neutronapis.yaml
+++ b/api/bases/neutron.openstack.org_neutronapis.yaml
@@ -2052,9 +2052,9 @@ spec:
                 default: memcached
                 description: Memcached instance name.
                 type: string
-              ml2Drivers:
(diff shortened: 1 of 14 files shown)


controllers/neutronapi_controller.go
```diff
@@ -922,6 +923,14 @@ func (r *NeutronAPIReconciler) reconcileNormal(ctx context.Context, instance *ne
 		return ctrl.Result{}, err
 	}
 
+	if instance.IsOVNEnabled() {
+		err = r.reconcileExternalOVNSecrets(ctx, helper, instance, &secretVars)
```
booxter, line 927: maybe reconcile should be executed anyway - to clean up secrets if the driver was disabled?

test/functional/neutronapi_controller_test.go
```diff
@@ -1485,7 +1544,7 @@ var _ = Describe("NeutronAPI Webhook", func() {
 		)
 	})
 
-	When("A NeutronAPI instance is updated with unsupported fields", func() {
+	When("A NeutronAPI instance is configured", func() {
```
booxter, line 1547: should be reverted

api/bases/neutron.openstack.org_neutronapis.yaml
```diff
@@ -1938,6 +1938,11 @@ spec:
                 default: memcached
                 description: Memcached instance name.
                 type: string
+              ml2Drivers:
+                description: 'TODO: describe the option'
```
booxter, line 1942: should add a proper description

booxter:
recheck

booxter:
recheck

FIPS check failure, should be ok after https://github.com/openshift/release/pull/57159

booxter:
`--- PASS: kuttl/harness/neutron_no_ovn (66.29s)` 🥳

booxter:
@karelyatin @slawqo I think this should be ready for review now. Thanks!

api/bases/neutron.openstack.org_neutronapis.yaml
```diff
@@ -2052,6 +2052,12 @@ spec:
                 default: memcached
                 description: Memcached instance name.
                 type: string
+              ml2Drivers:
```
slawqo, line 2055: shouldn't we name it like e.g. "ml2MechanismDrivers" as there are different types of drivers in ML2 plugin (type drivers, mechanism drivers, extension drivers for example)

booxter:
@karelyatin do you have any input here before I respin for what Slawek asked? (Renaming the field.)

api/v1beta1/neutronapi_types.go
```diff
@@ -102,6 +102,10 @@ type NeutronAPISpecCore struct {
 	// PreserveJobs - do not delete jobs after they finished e.g. to check logs
 	PreserveJobs bool `json:"preserveJobs"`
 
+	// +kubebuilder:validation:Optional
+	// Ml2Drivers - list of ml2 drivers to enable. Using ["ovn"] if not set.
```
karelyatin, line 106: wondering if can't we set default here itself or as part of webook instead of setting in GetMl2Drivers?

booxter:
@karelyatin @slawqo In the latest update, I:

- renamed the ml2Drivers field into ml2MechanismDrivers;
- implemented defaulting using CRD kubebuilder to fall back to ovn.

PR #410, opened 2024-09-10
refactor: Clean up some defaults from neutronapi config

booxter:
These are already the default values in upstream code.


karelyatin:
For kuttl failure needs https://github.com/openstack-k8s-operators/openstack-operator/pull/1065

PR #413, opened 2024-09-20
Run neutron httpd sidecar container using NeutronUID

fmount:
This patch moves the `httpd` sidecar container under `kolla` management and by doing this we are able to not run it as root anymore, but use the same `NeutronUID`/`NeutronGID` provided for the `NeutronAPI` service.

Part of the conversation is tracked in Jira [1].

Jira: https://issues.redhat.com/browse/OSPRH-10290

[1] https://issues.redhat.com/browse/OSPRH-2240

### DIFF
--- a/pkg/neutronapi/deployment.go
+++ b/pkg/neutronapi/deployment.go
@@ -30,8 +30,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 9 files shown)


abays (approved):
/lgtm

PR #421, opened 2024-10-08
Do not allow arbitary hostPath mounts via spec.extraMounts

karelyatin:
We are not using hostmount-anyuid scc, but having
it added in neutron rbac role was allowing
arbitary hostPath volume mounts via spec.extraMounts.

Removing hostmount-anyuid scc in order to not allow arbitary hostPath volume mounts and avoid unnecessary security hole[1].

[1] https://kubernetes.io/docs/concepts/storage/volumes/#hostpath
Related-Issue: OSPRH-10288
Closes-Issue: OSPRH-4297

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -235,7 +235,6 @@ rules:
   - security.openshift.io
   resourceNames:
   - anyuid
-  - hostmount-anyuid
   resources:
   - securitycontextconstraints
   verbs:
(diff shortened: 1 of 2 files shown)


booxter:
/lgtm

I agree with the direction (to forbid it instead of allowing it), but please update the linked story in Jira to reflect that, so that a reader in the future is not confused by the mismatch.

PR #428, opened 2024-10-18
[tls] Fix validate ovn service cert secret

Deydra71:
Add OVN cert validation to detect changes in `cert-neutron-ovndbs`. This patch includes the OVN cert hash in the combined TLS secrets hash.

### DIFF
--- a/controllers/neutronapi_controller.go
+++ b/controllers/neutronapi_controller.go
@@ -441,7 +441,7 @@ func (r *NeutronAPIReconciler) reconcileInit(
 	//
 	// Validate the CA cert secret if provided
 	if instance.Spec.TLS.CaBundleSecretName != "" {
-		hash, err := tls.ValidateCACertSecret(
+		caHash, err := tls.ValidateCACertSecret(
 			ctx,
 			helper.GetClient(),
 			types.NamespacedName{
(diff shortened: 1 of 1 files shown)


controllers/neutronapi_controller.go
```diff
@@ -491,7 +491,43 @@ func (r *NeutronAPIReconciler) reconcileInit(
...
+			secretVars["ovnCertsHash"] = env.SetValue(ovnCertsHash)
+		}
+	}
+
+	// Combine all cert hashes into a single hash
+	combinedCertsHash, err := util.HashOfInputHashes(secretVars)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	secretVars[tls.TLSHashName] = env.SetValue(combinedCertsHash)
```
stuggi, line 530: I don't think we need this. bellow in L539 there is a call for `r.createHashOfInputHashes()` which calcs the hash of hashes

Deydra71: Right, I thought about keeping the original logic with `tls.TLSHashName` but it's redundant to calculate hashes two times.

stuggi (approved):
/lgtm

PR #431, opened 2024-10-21
Improve automated support for local webhooks

abays:
Reworks `make run-with-webhook` target to trap `SIGINT` and `SIGTERM` so as to call `clean_local_webhook.sh` in such a scenario.  This way, a user can run the operator locally and automatically have the local webhooks removed when they finish running the operator via `ctrl+c`.


karelyatin (commented):
/lgtm

PR #432, opened 2024-10-25
Remove vxlan config bits from the default configuration

slawqo:
As we don't support vxlan network type in the ML2/OVN backend in RHOSO we should not load by default vxlan type driver. This patch cleans it from the default Neutron config template.

Closes: #[OSPRH-10967](https://issues.redhat.com//browse/OSPRH-10967)


slawqo:
recheck

PR #436, opened 2024-10-30
Make apache timeout configurable

karelyatin:
Added a parameter 'apiTimeout' to allow customization
to the Apache timeout.
The default is set to 120s which we do set for
HAProxy timeouts currently.

To be able to change the HAProxy value based on the `apiTimeout` with
any update (and not just the first time) the code adds a custom
annotation "api.neutron.openstack.org/timeout" with the value that was
initially set, this way flags it as being set by the neutron-operator.

There will be follow up patch in openstack-operator
to utilize the method 'SetDefaultRouteAnnotations' to set
these default route annotations.

Resolves: [OSPRH-10843](https://issues.redhat.com//browse/OSPRH-10843)

### DIFF
--- a/api/bases/neutron.openstack.org_neutronapis.yaml
+++ b/api/bases/neutron.openstack.org_neutronapis.yaml
@@ -48,6 +48,11 @@ spec:
           spec:
             description: NeutronAPISpec defines the desired state of NeutronAPI
             properties:
+              apiTimeout:
+                default: 120
+                description: APITimeout for HAProxy, Apache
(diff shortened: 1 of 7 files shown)


slawqo (approved):

averdagu:
Looks good to me! Thanks yatin

fernandoroyosanchez:
Looks good to me, doing something similar here https://github.com/openstack-k8s-operators/octavia-operator/pull/403

PR #437, opened 2024-11-06
Update to new lib-common VolumeSource structs

fmount:
This patch bumps `lib-common/storage` module and adopt the new `VolumeSource` struct required by `extraMounts`.


pkg/neutronapi/volumes.go
```diff
@@ -31,7 +31,14 @@ func GetVolumes(name string, extraVol []neutronv1beta1.NeutronExtraVolMounts, sv
 	}
 	for _, exv := range extraVol {
 		for _, vol := range exv.Propagate(svc) {
-			res = append(res, vol.Volumes...)
+			for _, v := range vol.Volumes {
+				volumeSource, _ := v.VolumeSource.ToCoreVolumeSource()
```
bshephar, line 35: It doesn't really matter, but the `VolumeSource` part of this is already implied, so it doesn't need to be explicitly mentioned. It could be:
```
v.ToCoreVolumeSource()
```

fmount: Done

abays (approved):
/lgtm

PR #450, opened 2024-12-09
Fix default cache config for no tls

karelyatin:
When env is deployed with tls disabled, cache config was wrong as "dogpile.cache.memcached" backend requires inet[6] prefixes for memcache_servers setting while it was set without the prefix. with IPv4 issue not observed as "inet" is default prefix, this patch fixes it by setting [cache]/memcache_servers based on tls config.

Resolves: OSPRH-12223
Related-Issue: OSPRH-12221


slawqo (approved):

stuggi (commented):
/lgtm

karelyatin:
quay.rdo io timeout

PR #463, opened 2025-01-13
Enable ovn_emit_need_to_frag by default

karelyatin:
The default switched in neutron master already with [1]. Setting it in operator default config to get it applied in older releases like antelope where [1] is not available.

[1] https://review.opendev.org/c/openstack/neutron/+/930608

Related-Issue: #[OSPRH-10684](https://issues.redhat.com//browse/OSPRH-10684)


ralonsoh:
LGTM

PR #465, opened 2025-01-16
[kuttl] Use relative path to scripts

karelyatin:
Couple of tests were relying on NEUTRON_KUTTL_DIR to be set, let's avoid the dep on extra var and use relative path to the test.

Alternative to https://github.com/openstack-k8s-operators/install_yamls/pull/915


karelyatin:
kuttl issue fixed

PR #471, opened 2025-02-07
Consume Topology CR by reference

fmount:
This patch provides more granular control over Pod placement and scheduling through the Topology CR integration introduced in [1]. In particular it provides:

- a new API parameter (TopologyRef) defined for each Component that allows to reference an existing Topology CRs in the same namespace

- the operator logic that retrieves and processes the referenced Topology CR through the functions provided by lib-common [2]

- enhanced StatefulSet and Deployment configuration that incorporates the processed Topology

- a set of envTest to test the lifecycle (add/update/override/remove) of the resulting StatefulSets when a Topology is referenced

Note that webhooks are in place to prevent referencing a Topology from a different namespace (which is not a supported scenario).

Jira: https://issues.redhat.com/browse/OSPRH-13786


danpawlik:
it will fail, due https://review.rdoproject.org/r/c/config/+/56290 needs to be merged before.

stuggi (approved):
/lgtm

PR #475, opened 2025-02-20
Update LastAppliedTopology to TopologyRef interface

fmount:
This patch updates the SpecCore webhook interface to accept the Namespace as parameter. This change ensures that we properly validate the referenced topology from openstack-operator, rejecting requests with a different namespace. It also aligns "LastAppliedTopology" to "TopologyRef" interface. While TopologyRef is able to reference Name and namespace, we currently only save TopologyRef.Name to .Status.LastAppliedTopology. This patch aligns both interfaces to use the same type in the same form. In addition, some logic is moved to infra-operator because is common to all operators, and this patch also adopts this new form.


stuggi:
@fmount I think this ci error is valid

```
# github.com/openstack-k8s-operators/openstack-operator/apis/core/v1beta1
apis/core/v1beta1/openstackcontrolplane_webhook.go:272:66: not enough arguments in call to r.Spec.Neutron.Template.ValidateCreate
	have (*"k8s.io/apimachinery/pkg/util/validation/field".Path)
	want (*"k8s.io/apimachinery/pkg/util/validation/field".Path, string)
apis/core/v1beta1/openstackcontrolplane_webhook.go:432:89: not enough arguments in call to r.Spec.Neutron.Template.ValidateUpdate
	have ("github.com/openstack-k8s-operators/neutron-operator/api/v1beta1".NeutronAPISpecCore, *"k8s.io/apimachinery/pkg/util/validation/field".Path)
	want ("github.com/openstack-k8s-operators/neutron-operator/api/v1beta1".NeutronAPISpecCore, *"k8s.io/apimachinery/pkg/util/validation/field".Path, string)
```

stuggi:
ah wait thats the one we talked about

fmount:
recheck

fmount:
recheck

stuggi (approved):
/lgtm

PR #506, opened 2025-05-19
Change memcache to use memcache_pool driver

stuggi:
In HA tests it was identified that pymemcache backend does not recover/fail over to the next memcache instance when one goes away unexpected. Using memcache_pool with bmemcache does failover properly.

The memcache_sasl_enable=true is a workaround to force oslo.cache to switch to the bmemcache driver, which is the only driver in memcache_pool supporting tls+fips.

Jira: [OSPRH-16651](https://issues.redhat.com//browse/OSPRH-16651)

Co-outhored-by: Luca Miccini <lmiccini@redhat.com>

### DIFF
--- a/templates/neutronapi/config/01-neutron.conf
+++ b/templates/neutronapi/config/01-neutron.conf
@@ -92,18 +92,21 @@ lock_path = /var/lib/neutron/tmp
 
 [cache]
 {{if .MemcachedTLS}}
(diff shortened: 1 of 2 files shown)


lmiccini (approved):

stuggi:
recheck

stuggi:
recheck

PR #514, opened 2025-06-06
Enable MTLS memcached auth

lmiccini:
This commit allows operators to use mtls as an authentication method against Memcached.
Neutron controller will detect the presence of a purposely-created mtls certificate (authCertSecret) and use this to configure the tls_context accordingly.
Additional volumes/volumemounts will be appended to each pod.


templates/neutronapi/config/01-neutron.conf
```diff
@@ -54,7 +54,7 @@ ovn_sb_ca_cert = /etc/pki/tls/certs/ovndbca.crt
 [keystone_authtoken]
 www_authenticate_uri = {{ .KeystonePublicURL }}
 auth_url = {{ .KeystoneInternalURL }}
-memcached_servers={{ .MemcachedServersWithInet }}
+memcached_servers={{ .MemcachedServers }}
```
karelyatin, line 57: Is this change intended? will this works with ipv6 setups withot inet6: prefix?

karelyatin (changes requested):

templates/neutronapi/config/01-neutron.conf
karelyatin: yes please do and also link the change which changed the behavior for context as we had regressions in past due to these, also will this work with tls disabled and ipv6?

karelyatin: Also https://review.opendev.org/c/openstack/keystonemiddleware/+/942202 is not backported to antelope

karelyatin: I see you reverted the patch mentioned in commit message https://review.opendev.org/c/openstack/keystonemiddleware/+/953169 , can you please update as per latest state

PR #517, opened 2025-06-17
Extend CRDs to support a RabbitMqCluster reference as AMQP instance for

auniyal61:
notifications.

To provide a way for human user to define the notifications bus for Neutronapi CR via the OpenStackControlPlane CR's neutron section, add a new field to the NeutronAPISpecCore as that is the struct used when the OpenStackControlPlaneSpec struct is defined.

At NeutronSpec level having just "" or a name of the bus is OK as either we need a bus, or we don't need notifications at all.

     “” - disabled,
     "foo" - external RabbitMQ Bus

Assume that in the implementation the required information will be transferred via the existing internal secret via notifications specific transport_url key (as the transport_url key is already taken by the RPC bus value).


auniyal61:
This is referenced from @bogdando original notification design for nova 
PR https://github.com/openstack-k8s-operators/nova-operator/pull/948

controllers/neutronapi_controller.go
```diff
@@ -928,6 +928,78 @@ func (r *NeutronAPIReconciler) reconcileNormal(ctx context.Context, instance *ne
 		return ctrl.Result{RequeueAfter: time.Duration(10) * time.Second}, nil
 	}
 
+	//
+	// notifications transporturl
+	//
+	notificationBusName := "rabbitmq-br"
```
auniyal61, line 934: nit: `notificationBusName := ""`
this was to local testing, will fix in next PS

auniyal61:
recheck deployment failed during kuttl test, unrelated to change 
`error: the server doesn't have a resource type "openstackversion"`

controllers/neutronapi_controller.go
```diff
@@ -928,6 +925,74 @@ func (r *NeutronAPIReconciler) reconcileNormal(ctx context.Context, instance *ne
...
+				err.Error()))
+			return result, err
+		} else if (result != ctrl.Result{}) {
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				condition.InputReadyCondition,
+				condition.RequestedReason,
+				condition.SeverityInfo,
+				condition.InputReadyWaitingMessage))
+			return result, err
+		}
```
stuggi, line 985: I don't think we have to verify the secret if we reached this point and `notificationTransportURL.IsReady() == true`, or am I missing condition ?

templates/neutronapi/config/01-neutron.conf
```diff
@@ -119,3 +119,12 @@ memcache_dead_retry = 30
 policy_file = /etc/neutron/policy.yaml
 enforce_scope = True
 enforce_new_defaults = True
+
+[oslo_messaging_notifications]
+{{- if .NotificationsTransportURL }}
+driver = messagingv2
+transport_url = {{ .NotificationsTransportURL }}
+# topics = notifications
```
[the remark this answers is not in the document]
stuggi: can not comment on this. would have to defer to @karelyatin or someone else from neutron

controllers/neutronapi_controller.go
```diff
@@ -1245,21 +1310,30 @@ func (r *NeutronAPIReconciler) reconcileNormal(ctx context.Context, instance *ne
...
+func (r *NeutronAPIReconciler) transportURLCreateOrUpdate(
+	ctx context.Context,
+	instance *neutronv1beta1.NeutronAPI,
+	rabbitmqName string) (
+	*rabbitmqv1.TransportURL, controllerutil.OperationResult, error) {
+	Log := r.GetLogger(ctx)
 	transportURL := &rabbitmqv1.TransportURL{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-neutron-transport", instance.Name),
+			Name:      fmt.Sprintf("neutron-%s-transport", rabbitmqName),
```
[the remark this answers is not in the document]
auniyal61: > how would we do the cleanup of the current existing transporturl for envs which get updated?

added cleanup

> if we change the name, we get a new transport url also for the non notification bus

yes
> neutron will restart because it gets a new transport url, even if notifiation bus feature is not enabled.

greenfield should be fine,
yes, it will affect upgrade/minor-update otherwise operator have to delete manually - will look into this.

controllers/neutronapi_controller.go
```diff
@@ -928,6 +925,74 @@ func (r *NeutronAPIReconciler) reconcileNormal(ctx context.Context, instance *ne
...
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				condition.NotificationBusInstanceReadyCondition,
+				condition.ErrorReason,
+				condition.SeverityWarning,
+				condition.NotificationBusInstanceReadyErrorMessage,
+				err.Error()))
+			return ctrl.Result{}, err
+		}
+
+		if notificationTransportURL.Status.SecretName == "" {
```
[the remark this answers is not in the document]
auniyal61: ack, though I wanted to follow the existing flow.

controllers/neutronapi_controller.go
auniyal61: yes, make sense.
can reemove  condition update for `condition.InputReadyErrorMessage`  and `condition.InputReadyWaitingMessage`
its been done for default transport-url too.

@karelyatin

controllers/neutronapi_controller.go
```diff
@@ -1393,6 +1467,23 @@ func (r *NeutronAPIReconciler) getTransportURL(
...
+) (string, error) {
+	transportURLSecret, _, err := secret.GetSecret(ctx, h, *instance.Status.NotificationsTransportURLSecret, instance.Namespace)
+	if err != nil {
+		return "", err
+	}
+
+	transportURL, ok := transportURLSecret.Data["transport_url"]
+	if !ok {
+		return "", fmt.Errorf("transport_url %w Transport Secret", util.ErrNotFound)
+	}
```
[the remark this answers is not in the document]
auniyal61: done

templates/neutronapi/config/01-neutron.conf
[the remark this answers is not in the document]
karelyatin: me neither know much about notifications so can't add much here, @slawqo @ralonsoh can tell better

controllers/neutronapi_controller.go
```diff
@@ -1245,24 +1335,58 @@ func (r *NeutronAPIReconciler) reconcileNormal(ctx context.Context, instance *ne
...
 		},
 	}
 
 	op, err := controllerutil.CreateOrUpdate(context.TODO(), r.Client, transportURL, func() error {
-		transportURL.Spec.RabbitmqClusterName = instance.Spec.RabbitMqClusterName
+		transportURL.Spec.RabbitmqClusterName = rabbitmqName
 
 		err := controllerutil.SetControllerReference(instance, transportURL, r.Scheme)
 		return err
 	})
```
karelyatin, line 1358: some debug statements including XXX can be dropped.

controllers/neutronapi_controller.go
```diff
@@ -928,6 +926,98 @@ func (r *NeutronAPIReconciler) reconcileNormal(ctx context.Context, instance *ne
...
+		transportURLList := &rabbitmqv1.TransportURLList{}
+		listOpts := []client.ListOption{
+			client.InNamespace(instance.Namespace),
+		}
+		if err := r.Client.List(ctx, transportURLList, listOpts...); err != nil {
+			return ctrl.Result{}, err
+		}
+
+		for _, url := range transportURLList.Items {
+			if strings.Contains(url.Name, "neutron") && strings.Contains(url.Name, "notifications") {
```
karelyatin, line 1007: since we generating the name above can use same name to detect existence and removal.

controllers/neutronapi_controller.go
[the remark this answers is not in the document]
karelyatin: +1 to what Martin said and if we can avoid the renames for existing transport URLs

controllers/neutronapi_controller.go
auniyal61: yes, we can say `"neutron-"+notificationBusName+"-notifications"+"-transport"`, but that is more hardcoding, 
finding from neutron-notification transport url and then deleting is better IMO.

controllers/neutronapi_controller.go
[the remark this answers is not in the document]
auniyal61: default transporturl name will set as `neutron-neutron-transport"` as earlier.
updated test as well.

templates/neutronapi/config/01-neutron.conf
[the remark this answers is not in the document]
auniyal61: I removed the topic for now, could not see appropriate upstream doc for, 
can be added if required in future

controllers/neutronapi_controller.go
```diff
@@ -1683,6 +1804,20 @@ func (r *NeutronAPIReconciler) generateServiceSecrets(
...
+
+		if err != nil {
+			return err
+		}
+
+		templateParameters["NotificationsTransportURL"] = notificationsTransportURL
+
+	} else {
+		templateParameters["NotificationsTransportURL"] = nil
+	}
```
stuggi, line 1819: do we even have to do this else case? from the template it seems we expect it has a good url if it is set.

controllers/neutronapi_controller.go
```diff
@@ -1683,6 +1804,20 @@ func (r *NeutronAPIReconciler) generateServiceSecrets(
 	templateParameters["MemcachedTLS"] = mc.GetMemcachedTLSSupport()
 	templateParameters["TimeOut"] = instance.Spec.APITimeout
 
+	if instance.Status.NotificationsTransportURLSecret != nil {
+		notificationsTransportURL, _, err := secret.GetDataFromSecret(
```
stuggi, line 1808: maybe we could update `getTransportURL()` to be also used for the notifications url. like passing in instance.Status.NotificationsTransportURLSecret as a parameter

controllers/neutronapi_controller.go
auniyal61: we need this to explicitly set to nil, because in conf templates we are checking if its value is set, 
if we do not set it, it get set to null and we get regression. for job db-sync fail

its same as setting for `IsOVN` at L1844.

I moved it above for better setting and readability.

controllers/neutronapi_controller.go
```diff
@@ -1381,8 +1505,12 @@ func (r *NeutronAPIReconciler) getTransportURL(
 	ctx context.Context,
 	h *helper.Helper,
 	instance *neutronv1beta1.NeutronAPI,
+	secretName *string,
 ) (string, error) {
-	transportURLSecret, _, err := secret.GetSecret(ctx, h, instance.Status.TransportURLSecret, instance.Namespace)
+	if secretName == nil || *secretName == "" {
+		return "", errTransportURLSecretNameNilOrEmpty
```
auniyal61, line 1511: //nolint:err113
I am also inclind towards ignoring lint for this specific case.

---
so reason we are getting pre-commit error is we do not have error obj to typecast here for fmt.Errorf like in L1519, and to convert string to error errors.New(""), have to move the declaration at top package level.
but to me this seems wrong/cluttering.

controllers/neutronapi_controller.go
```diff
@@ -69,6 +70,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// errTransportURLSecretNameNilOrEmpty
+var errTransportURLSecretNameNilOrEmpty = errors.New("transport_url secret name is nil or empty")
+
```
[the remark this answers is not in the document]
stuggi: I think its fine to what it is right now. What we in general could do is to move the transport url consumer funcs (transportURLCreateOrUpdate/transportURLDeleted/getTransportURL) to https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/rabbitmq/v1beta1/transporturl_types.go and all service operators could then consume them with this static err

auniyal61:
Note for how kuttl tests finally worked
during creation of new rabbit reduce resource consumptions

```
  persistence:
    storage: 1Gi
  resources:
    limits:
      cpu: "1"
      memory: 1Gi
    requests:
      cpu: "0.5"
      memory: 500Mi
```

also the rabbit definition

test/kuttl/tests/notifications/03-assert.yaml
```diff
@@ -0,0 +1,7 @@
+# Verify notifications transporturl was cleaned up (command should fail)
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+  - script: |
+      oc get transporturl/notifications-neutron-transport -n $NAMESPACE || exit 0
```
stuggi, line 6: iiuc you want to check that the notirifations-neutron-transport cr is gone. you could do this using an errors file, like https://github.com/openstack-k8s-operators/openstack-operator/blob/main/tests/kuttl/tests/ctlplane-tls-custom-route/04-errors-cleanup.yaml, to check that a resource is gone and you expect an error.

stuggi:
other then the comment on kuttl it looks good to me

test/kuttl/tests/notifications/02-assert.yaml
```diff
@@ -0,0 +1,8 @@
+# Verify external notification config
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+  - script: |
+        oc wait --for=condition=Ready -n $NAMESPACE pod -l service=neutron --timeout=300s
```
karelyatin, line 6: duplicated with 01-assert

controllers/neutronapi_controller.go
```diff
@@ -1682,6 +1806,12 @@ func (r *NeutronAPIReconciler) generateServiceSecrets(
 	templateParameters["MemcachedServersWithInet"] = mc.GetMemcachedServerListWithInetString()
 	templateParameters["MemcachedTLS"] = mc.GetMemcachedTLSSupport()
 	templateParameters["TimeOut"] = instance.Spec.APITimeout
+	templateParameters["NotificationsTransportURL"] = nil
```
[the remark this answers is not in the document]
auniyal61: paste the same reply as earlier.

we need this to explicitly set to nil, because in conf templates we are checking if its value is set,
if we do not set it, it get set to null and we get regression. for job db-sync fail

its same as setting for IsOVN at L1844.

controllers/neutronapi_controller.go
```diff
@@ -1682,6 +1808,16 @@ func (r *NeutronAPIReconciler) generateServiceSecrets(
...
+
+	notificationsTransportURL, err := r.getTransportURL(ctx, h, instance, instance.Status.NotificationsTransportURLSecret)
+	if err != nil && !errors.Is(err, errTransportURLSecretNameNilOrEmpty) && !k8s_errors.IsNotFound(err) {
```
karelyatin, line 1814: why IsNotFound excluded?

controllers/neutronapi_controller.go
```diff
@@ -1658,9 +1780,11 @@ func (r *NeutronAPIReconciler) generateServiceSecrets(
 		return err
 	}
 
-	transportURL, err := r.getTransportURL(ctx, h, instance)
+	transportURL, err := r.getTransportURL(ctx, h, instance, &instance.Status.TransportURLSecret)
 	if err != nil {
-		return err
+		// If transport URL is not ready yet, we can't generate the main config
+		// but we should not return an error - instead wait for it on reconcile
```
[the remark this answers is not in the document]
karelyatin: error returned now

test/kuttl/tests/notifications/01-assert.yaml
```diff
@@ -0,0 +1,7 @@
+# Verify the RabbitMQ cluster rabbitmq-br created for notifications is Ready
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+  - script: |
+      oc get rabbitmqcluster.rabbitmq.com rabbitmq-br -n $NAMESPACE || exit 1
+      exit 0
```
auniyal61, line 7: will try with conditions

test/kuttl/tests/notifications/01-assert.yaml
```diff
@@ -0,0 +1,29 @@
+# Verify the RabbitMQ cluster rabbitmq-br created for notifications is Ready
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+  - script: |
+      oc wait --for=condition=ReconcileSuccess rabbitmqcluster.rabbitmq.com rabbitmq-br -n $NAMESPACE --timeout=300s
```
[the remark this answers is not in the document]
karelyatin: 5 minute too much for this, also can reduce timeout as well to less that 180 sec as thats set per test.

test/kuttl/tests/notifications/01-assert.yaml
```diff
@@ -0,0 +1,19 @@
...
+apiVersion: rabbitmq.com/v1beta1
+kind: RabbitmqCluster
+metadata:
+  name: rabbitmq-br
+status:
+  conditions:
+    - type: ClusterAvailable
+      status: "True"
+    - type: ReconcileSuccess
+      status: "True"
```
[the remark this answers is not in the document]
karelyatin: Above Testassert using wait is enough so can even remove the conditions check. So can just sync limits == requests to avoid these warning

test/kuttl/tests/notifications/02-assert.yaml
```diff
@@ -0,0 +1,24 @@
...
+    reason: Ready
+    status: "True"
+    type: TransportURLReady
+---
+# Verify external notification config
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+  - script: |
+        oc wait --for=condition=Ready -n $NAMESPACE pod -l service=neutron --timeout=300s
```
karelyatin, line 22: reduce timeout here as well

test/kuttl/tests/notifications/01-assert.yaml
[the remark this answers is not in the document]
auniyal61: done

auniyal61:

```
    logger.go:42: 15:32:04 | notifications/2-notifications-config | pod/neutron-78c588df55-9fk7w condition met
    logger.go:42: 15:32:04 | notifications/2-notifications-config | Defaulted container "neutron-api" out of: neutron-api, neutron-httpd
    logger.go:42: 15:32:10 | notifications/2-notifications-config | error: Internal error occurred: error executing command in container: container is not created or running
    logger.go:42: 15:32:11 | notifications/2-notifications-config | running command: [sh -c oc wait --for=condition=Ready -n $NAMESPACE pod -l service=neutron --timeout=180s
        neutron_pod=$(oc get pods -n $NAMESPACE -l service=neutron --field-selector=status.phase=Running -o name | head -1)
        oc rsh -n $NAMESPACE ${neutron_pod} crudini --get /etc/neutron/neutron.conf.d/01-neutron.conf oslo_messaging_notifications driver | grep -q messagingv2
        ]
    logger.go:42: 15:35:11 | notifications/2-notifications-config | timed out waiting for the condition on pods/neutron-5fd885999-2zjl7
    logger.go:42: 15:35:11 | notifications/2-notifications-config | timed out waiting for the condition on pods/neutron-78c588df55-9fk7w
    logger.go:42: 15:35:11 | notifications/2-notifications-config | test step failed 2-notifications-config
    case.go:378: failed in step 2-notifications-config


    case.go:380: command "oc wait --for=condition=Ready -n $NAMESPACE pod -l service=neutron ..." exceeded 141 sec timeout, context deadline exceeded
    logger.go:42: 15:35:11 | notifications | skipping kubernetes event logging
=== CONT  kuttl/harness/neutron_scale
```

PR #523, opened 2025-07-01
Sort map keys for deterministic iteration

karelyatin:
The map ordering is random in golang[1], using
sort to make this deterministic and avoid any
reconcilation loops

[1] https://go.dev/blog/maps#iteration-order

Resolves: [OSPRH-17883](https://issues.redhat.com//browse/OSPRH-17883)


averdagu:
Looks good to me

PR #534, opened 2025-07-30
Add support for the neutron-fwaas plugin

slawqo:
This patch adds "enableFwaas" flag to the Neutron CRD. Setting this flag to `True` will enable `firewall_v2` service plugin and configure OVN service provider for the FIREWALL_V2 resource if OVN mechanism driver is used.

Closes: #[OSPRH-15214](https://issues.redhat.com//browse/OSPRH-15214)

### DIFF
--- a/api/bases/neutron.openstack.org_neutronapis.yaml
+++ b/api/bases/neutron.openstack.org_neutronapis.yaml
@@ -89,6 +89,10 @@ spec:
                 description: DefaultConfigOverwrite - interface to overwrite default
                   config files like policy.yaml
                 type: object
+              enableFwaas:
+                default: false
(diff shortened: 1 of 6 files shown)


slawqo:
> if this is only setting config values, shouldn't this be a docs thing on how to use the custom service override to configure it? we intended to not add tht like parameters to just set config values. if there is deployment related changes required based on the switch its ok to add.

It is not handled in this PR yet but in case when neutron-fwaas was enabled in the existing deployment we should run neutron-db-manage script to update database and create fwaas tables. I think that this may be easier that way rather then parsing `customServiceConfig` and looking for the service_plugins there.

PR #541, opened 2025-08-19
Use quorum queues if enabled

lmiccini:
With https://github.com/openstack-k8s-operators/infra-operator/pull/429 we added the ability to configure RabbitMQ to use quorum queues.

This commit adds a check to verify if "quorumqueues" has been added to the TransportURL secret. If its value is true, the services will be configured to use quorum queues.

Jira: https://issues.redhat.com/browse/OSPRH-19160


controllers/neutronapi_controller.go
```diff
@@ -1513,6 +1513,22 @@ func (r *NeutronAPIReconciler) getTransportURL(
 	return string(transportURL), nil
 }
 
+func (r *NeutronAPIReconciler) isQuorumQueuesEnabled(
```
fmount, line 1516: @lmiccini should this be an helper function common to all operators? I'm wondering if this code can be centralized in infra-operator so that other services can access the same function in case we need it.
In addition, while I understand this is a very explicit check, I'm wondering if having:

```golang
transportURLQuorumQueues := string(transportURLSecret.Data["quorumqueues"]) == "true"
```
would be enough.

lmiccini: nice catch, it's probably a leftover, let me try with the oneliner (should work but I don't trust myself)

stuggi (approved):
/lgtm

PR #629, opened 2026-05-07
Add AGENTS.md

fmount:
This patch adds `AGENTS.md` to the repo to ensure any `AI agent` can follow the conventions defined for this operator. At the same time, it allows to not waste time navigating the repo to find initial information and save tokens.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,89 @@
+# AGENTS.md - neutron-operator
+
+## Project overview
+
+neutron-operator is a Kubernetes operator that manages
+[OpenStack Neutron](https://docs.openstack.org/neutron/latest/) (the networking
+service: virtual networks, routers, security groups, floating IPs, ML2/OVN,
+DHCP, metadata) on OpenShift/Kubernetes. It is part of the
+[openstack-k8s-operators](https://github.com/openstack-k8s-operators) project.
+
+Key Neutron domain concepts: **virtual networks**, **subnets**, **routers**,
+**security groups**, **floating IPs**, **ML2 plugin**, **OVN backend**,
+**DHCP agent**, **metadata agent**, **SR-IOV agent**, **OVN agent**.
+
+## Tech stack
+
+| Layer | Technology |
+|-------|------------|
+| Language | Go (modules, multi-module workspace via `go.work`) |
+| Scaffolding | [Kubebuilder v4](https://book.kubebuilder.io/) + [Operator SDK](https://sdk.operatorframework.io/) |
+| CRD generation | controller-gen (DeepCopy, CRDs, RBAC, webhooks) |
+| Config management | Kustomize |
+| Packaging | OLM bundle |
+| Testing | Ginkgo/Gomega + envtest (functional), KUTTL (integration) |
+| Linting | golangci-lint (`.golangci.yaml`) |
+| CI | Zuul (`zuul.d/`), Prow (`.ci-operator.yaml`), GitHub Actions |
+
+## Custom Resources
+
+| Kind | Purpose |
+|------|---------|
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,87 @@
...
+## Directory structure
+
+| Directory | Contents |
+|-----------|----------|
+| `api/v1beta1/` | CRD types (`neutronapi_types.go`), conditions, webhook markers |
+| `cmd/` | `main.go` entry point |
+| `internal/controller/` | Reconciler: `neutronapi_controller.go` |
+| `internal/neutronapi/` | NeutronAPI resource builders (deployment, volumes, config) |
+| `internal/webhook/` | Webhook implementation |
+| `templates/` | Config files mounted into pods via `OPERATOR_TEMPLATES` env var: `dhcp-agent.conf`, `ovn-agent.conf`, `ovn-metadata-agent.conf`, `sriov-agent.conf`, `neutronapi/` (config and httpd templates) |
```
karelyatin: the agent.conf files are basically used to create secrets which are later used by dataplane services to deploy on the edpm nodes.

AGENTS.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Testing
+
+- Functional tests use the envtest framework with Ginkgo/Gomega and live in
+  `test/functional/`.
+- KUTTL integration tests live in `test/kuttl/`.
+- Run all functional tests: `make test`.
+- When adding a new field or feature, add corresponding test cases in
+  `test/functional/` and update fixture data accordingly.
+
```
karelyatin, line 81: can also add about pre-commit what we run in CI to cover santiy/lint/crd-checker etc checks

AGENTS.md
karelyatin: also will this file be auto updated if some of the basic constructs are changed in the repo using the AI tool? or would require explicit instructions or need to add the instruction in AGENTS.md now?

AGENTS.md
```diff
@@ -0,0 +1,87 @@
...
+## Key dependencies
+
+- [lib-common](https://github.com/openstack-k8s-operators/lib-common): shared modules for conditions, endpoints, database, TLS, secrets, etc.
+- [infra-operator](https://github.com/openstack-k8s-operators/infra-operator): RabbitMQ and topology APIs.
```
karelyatin: can add memcached too here as we using it here.

AGENTS.md
```diff
@@ -0,0 +1,87 @@
...
+| `test/kuttl/` | KUTTL integration tests |
+| `hack/` | Helper scripts (CRD schema checker, local webhook runner) |
+
+## Build commands
+
+After modifying Go code, always run: `make generate manifests fmt vet`.
+
+## Code style guidelines
+
+- Follow standard openstack-k8s-operators conventions and lib-common patterns.
```
karelyatin: how it will know about the patterns used in don't we need to add some link here to those?

AGENTS.md
fmount: ack thanks, I will add a few words about this.

AGENTS.md
fmount: yes, makes sense, adding in my next iteration.

AGENTS.md
fmount: I'm not entirely sure. I think it would be better to explicitly have a rule to update this file in case the directory tree is altered or any relevant content mentioned in this file is modified.

AGENTS.md
fmount: I'm adding the links and include them in the key dependencies (including the developer guide that is technically a dependency for an AI assistant)

AGENTS.md
fmount: Done

slawqo (approved):

karelyatin (approved):
/lgtm

karelyatin (approved):
/lgtm

PR #669, opened 2026-08-06
Remove kolla entrypoint pattern, harden SecurityContext

stuggi:
Eliminate kolla_start and the config.json staging pattern across all neutron-operator workloads (neutron-api, neutron-httpd, db-sync). Configs are now mounted directly to final paths via SubPath volume mounts. neutron-operator already uses Pattern B (two self-contained Secrets: one for neutron-api/db-sync, one for neutron-httpd), like keystone/placement/glance — no conf.d-merge logic to reproduce.

This repo already tried and reverted a kolla-removal attempt in 2023 (7a750f10, reverted by 62a2f295). That attempt replaced kolla with a custom init-container needing sudo, which forced
AllowPrivilegeEscalation: true and blocked dropping ALL capabilities — a worse security posture than kolla itself. This migration does not repeat that mistake: all config rendering happens server-side in the Go controller, Secrets are mounted read-only via SubPath, and no runtime script/sudo is involved anywhere, the same proven approach already used for keystone/placement/cinder/glance.

Key changes:
- Replace kolla_start with direct commands per container/job (neutron-server, httpd -DFOREGROUND, neutron-db-manage) taken straight from each removed config.json's "command" field; the db-sync Job previously had no explicit Command/Args at all, relying on the image's default entrypoint — now explicit like every other workload
- Remove KOLLA_CONFIG_STRATEGY env var and the 3 kolla config.json templates
- Mount config-data keys via SubPath directly to /etc/neutron/neutron.conf.d/*.conf, /etc/httpd/conf(.d)/*, /etc/my.cnf; policy.yaml is mounted conditionally, only when set via spec.defaultConfigOverwrite (an unconditional mount would break any NeutronAPI that doesn't set it, since the secret key wouldn't exist)
- BUG FIX: Add DefaultMode 0440 to both the "config" and "httpd-config" Secret volumes. Neither had a DefaultMode set, meaning k8s defaulted to 0644 (world-readable). These Secrets contain DB credentials, transport URLs, and service passwords that should only be readable by the service user and group — not by every process in the container
- Adopt lib-common volume helpers (volume.WritableDirVolume, volume.WritableDirVolumeMount) with named constants (volume.RunHttpdVolumeName, volume.RunHttpdMountPath) for the run-httpd emptyDir, replacing inline constructions; removes the local GetHttpdRunVolume() function
- Fix memcached mTLS cert/key mounting: CreateMTLSVolumeMounts(nil, nil) mounted at the staging default path instead of where the rendered config actually references them (same bug already found and fixed in cinder-operator, keystone-operator, and glance-operator)
- Fix the OVN DB TLS cert mounts: the tls.Service literal only set CaMount, hardcoded to a kolla staging path even though 01-neutron.conf's ovn_nb_ca_cert/ovn_sb_ca_cert already pointed at the final /etc/pki/tls/certs/ovndbca.crt destination; CertMount/ KeyMount were left unset entirely, so the client cert/key (ovn_{nb,sb}_certificate/private_key) defaulted to a different kolla staging path with nothing there, crashing neutron-server at ML2/OVN driver init with FileNotFoundError on ovndb.crt. Found via real-cluster testing -- no functional or kuttl test exercises Spec.TLS.Ovn at all, so this was invisible to `make test`
- Use pod.RestrictiveSecurityContext/RestrictivePodSecurityContext from lib-common for all three workloads — no privileged/host-access exception needed anywhere in this repo (unlike cinder/glance)
- Set AutomountServiceAccountToken=false on all workloads
- Migrate SCC rbacRules from [anyuid] to [nonroot-v2] only — no privileged fallback needed anywhere in this repo, the first operator in this effort with that simpler end state
- Update functional and kuttl tests

Jira: [OSPRH-33504](https://redhat.atlassian.net/browse/OSPRH-33504)    
Jira: [OSPRH-33503](https://redhat.atlassian.net/browse/OSPRH-33503)    

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/728


stuggi:
recheck

stuggi:
recheck

abays (approved):
/lgtm

